PR #7, opened 2018-10-05
meta-mender-intel: initial template files for Minnowboard

mirzak:

### DIFF
--- a/meta-mender-intel/scripts/manifest-intel.xml
+++ b/meta-mender-intel/scripts/manifest-intel.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+  <manifest>
+  <default sync-j="4" revision="sumo"/>
+
+  <remote fetch="git://git.yoctoproject.org"        name="yocto"/>
+
+  <project name="poky" remote="yocto" revision="sumo" path="sources/poky"/>
+  <project name="meta-intel" remote="yocto" revision="sumo" path="sources/meta-intel"/>
+
+  <include name="scripts/mender.xml"/>
+
+  </manifest>
+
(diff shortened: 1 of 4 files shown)


mirzak:
@estenberg, this is what the Minnowboard would look like here. Based on your instructions in Mender Hub.

I have build tested this.

meta-mender-intel/scripts/manifest-intel.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+  <manifest>
+  <default sync-j="4" revision="sumo"/>
+
+  <remote fetch="git://git.yoctoproject.org"        name="yocto"/>
+  <remote fetch="git://github.com/linux-sunxi"      name="sunxi"/>
```
estenberg: Is this needed?

meta-mender-intel/scripts/manifest-intel.xml
```diff
@@ -0,0 +1,16 @@
...
+  <remote fetch="git://git.openembedded.org"        name="oe"/>
```
estenberg: Why do we need openembedded? Isn't yoctoproject enough?

meta-mender-intel/scripts/manifest-intel.xml
mirzak: No, leftover from copy/paste

meta-mender-intel/scripts/manifest-intel.xml
mirzak: Also a left over from copy/paste. Poky is enough and meta-intel does not have any other dependencies.

mirzak:
> I suppose the Mender Hub Minnowboard page should be updated?

Yes once this is merged and validated we should update  Minnowboard page.

> Moreover, should we update the template as well?

Yes, I think it is a good idea.

mirzak:
Updated, also note one more addition to the `local.conf.append`. I had some issues building `core-image-base` and this tries to address that.

estenberg:
Looks good, thanks for fixing @mirzak !
Would you mind updating the minnowboard topic in hub as well?

mirzak:
I would not mind :)

mirzak:
Updated the minnowboard topic

PR #8, opened 2018-10-09
Added support for beaglebone boards

nandra:
Signed-off-by: Marek Belisko <marek.belisko@open-nandra.com>


meta-mender-beaglebone/templates/local.conf.append
```diff
@@ -0,0 +1,7 @@
+
+# Appended fragment from meta-mender-community/meta-mender-beaglebone/templates
+
+MACHINE ?= "beaglebone-yocto"
+
+MENDER_BOOT_PART_SIZE_MB = "40"
+IMAGE_INSTALL_append = " kernel-image kernel-devicetree"
```
mirzak: This is not needed as it is set in `beaglebone-yocto.conf`.

meta-mender-beaglebone/templates/local.conf.append
```diff
@@ -0,0 +1,7 @@
+
+# Appended fragment from meta-mender-community/meta-mender-beaglebone/templates
+
+MACHINE ?= "beaglebone-yocto"
+
+MENDER_BOOT_PART_SIZE_MB = "40"
```
mirzak: Did you have issues with the default values for above? Would not expect this to be required on BBB.

scripts/setup-environment
```diff
@@ -23,7 +23,8 @@ targets=(
     "renesas"
     "rockchip"
     "sunxi"
-    "variscite")
+    "variscite"
+    "beaglebone")
```
mirzak: Lets try to keep this list in alphabetical order.

meta-mender-beaglebone/templates/local.conf.append
mirzak: But if you want to keep it does not hurt either. Up to you.

nandra:
Thanks for review. I'll push fixup shortly (it's ok with commit on top or should I squash?). thx

mirzak:
Squash and force push is prefererad.

mirzak:
Seems like a "merge commit" sneaked in :)

mirzak:
I will clean that up so do no worry about it

nandra:
Yes something like that but resolved already ;)

mirzak:
Yeah what mean is that you did a "git merge sumo" right? And you now got an "extra commit" saying that you did that which will follow your other commit if I was to merge this.

The more "clean" way would be to run "git rebase sumo", if there where changes on sumo that you want to include in your branch to avoid having "merge commits" in your PR.

nandra:
Oh yes right. I'll do it.

nandra:
Not it should be 100% :100:  :)

PR #9, opened 2018-10-29
Added support for orangepi zero

nandra:
Few patches which add wifi support, but one blocking point (kernel doesn't boot without this change: https://github.com/linux-sunxi/meta-sunxi/issues/232)

### DIFF
--- a/meta-mender-sunxi/recipes-kernel/linux/linux-mainline_%.bbappend
+++ b/meta-mender-sunxi/recipes-kernel/linux/linux-mainline_%.bbappend
@@ -1,0 +1,4 @@
+do_kernel_configme[depends] += "virtual/${TARGET_PREFIX}binutils:do_populate_sysroot"
+do_kernel_configme[depends] += "virtual/${TARGET_PREFIX}gcc:do_populate_sysroot"
+do_kernel_configme[depends] += "bc-native:do_populate_sysroot bison-native:do_populate_sysroot"
+


nandra:
Not sure if wifi driver + fw should be here or move to meta-sunxi

mirzak:
This is definitely something that I would target at meta-sunxi instead.

I think it would be OK to add specific patches here, "for staging" if they are patches required to be able to work with Mender or to event boot the device. WiFi functionality seems to be on the edge here as the orangepi zero seems to have an ethernet interface as well, so this is "optimization". Do you think this makes sense?

Regarding zImage problems, I experienced the opposite problems with NanoPi and Orange Pi PC Plus, where zImage worked for these but not uImage.

There is even a change in the templates for this, https://github.com/mendersoftware/meta-mender-community/blob/sumo/meta-mender-sunxi/templates/local.conf.append#L8-L10

If uImage works you can probably update the `local.conf.append` with machine specific overrides instead

nandra:
Ok I'll push that to meta-sunxi instead. orange-pi-zero is using mainline 4.16 kernel nd both uImage + zImage works fine. So I'll update this PR accordingly. Thanks.

mirzak:
@nandra, how did it go with pushing this to meta-sunxi?

nandra:
@mirzak it was accepted and merged just yesterday. I'll update PR accordingly. Thanks.

nandra:
OK branch updated with small single change. So for sumo wifi will not work. It will work for thud (but thud not yet pushed in meta-sunxi). Thx.

mirzak:
I will merge this, but can you create a PR on meta-sunxi/sumo with the same changes? Makes sense to add it there, but we can keep it around until then.

nandra:
OK not sure if maintainer allow that but I'll try. Thanks.

mirzak:
Well it fixes a build error? They should :).

Is there a Mender Hub post coming up for this as well? :D

nandra:
Oh this one. Yes they have it already in master I'll try to push it to sumo also. For mender hub -> sure thing :+1:

nandra:
Done: https://github.com/linux-sunxi/meta-sunxi/pull/241

nandra:
Done: https://hub.mender.io/t/orangepi-zero/142

PR #21, opened 2018-11-28
Compulab updates

drewmoseley:


mirzak:
Why the removal om meta-mender-demo? Generally I think it should be including here

drewmoseley:
Is that our standard here? That seems to be pretty invasive and usable only for the demo server environment and not hosted.

mirzak:
Yes, this has been so far the standard. It is after all reference images to quickly test Mender, and my impression (based on discussions on mender-convert) it is ok to have demo images with demo values pointed to Hosted Mender.

There is a pending discussion/task on me to include server information to Mender Hub posts, where we might evaluate this if needed.

drewmoseley:
OK.  I removed the commit so that meta-mender-demo is included now by default. This is ready for merge.

PR #23, opened 2018-12-04
Setup display resolution and touch panel on Colibri i.MX7

drewmoseley:
This is to address https://tracker.mender.io/browse/MEN-2057


mirzak:
Nice, looks good

PR #35, opened 2018-12-10
meta-sunxi: Thud updates

nandra:


meta-mender-sunxi/recipes-bsp/u-boot/files/0002-env-Kconfig-remove-defaults-for-SUNXI.patch
```diff
@@ -1,43 +1,36 @@
...
--	hex "Environment Offset"
--	depends on !ENV_IS_IN_UBI
--	depends on !ENV_IS_NOWHERE
--	default 0x88000 if ARCH_SUNXI
--	help
--	  Offset from the start of the device (or partition)
--
- config ENV_SIZE
- 	hex "Environment Size"
+@@ -438,7 +437,6 @@ config ENV_OFFSET
```
mirzak: Can not leave this one unfortunately because it will default to ENV_OFFSET="", and as you might have noticed that meta-mender complains about this.

meta-mender-sunxi/recipes-bsp/u-boot/files/0003-Enable-env-is-in-mmc-define-offset.patch
```diff
@@ -0,0 +1,32 @@
...
+ 1 file changed, 2 insertions(+)
+
+diff --git a/configs/orangepi_zero_defconfig b/configs/orangepi_zero_defconfig
+index 0da9c65..2b585c9 100644
+--- a/configs/orangepi_zero_defconfig
++++ b/configs/orangepi_zero_defconfig
+@@ -1,5 +1,6 @@
+ CONFIG_ARM=y
+ CONFIG_ARCH_SUNXI=y
++CONFIG_ENV_OFFSET=0x800000
```
mirzak: This is due to the fact that "config ENV_OFFSET" was left visible. But this actually problematic as meta-mender is supposed to set ENV_OFFSET based which partition alignment you chose.

meta-mender-sunxi/recipes-bsp/u-boot/files/0003-Enable-env-is-in-mmc-define-offset.patch
```diff
@@ -0,0 +1,32 @@
...
+ CONFIG_SPL=y
+ CONFIG_MACH_SUN8I_H3=y
+ CONFIG_DRAM_CLK=624
+@@ -12,6 +13,7 @@ CONFIG_NR_DRAM_BANKS=1
+ CONFIG_CONSOLE_MUX=y
+ # CONFIG_CMD_FLASH is not set
+ CONFIG_DEFAULT_DEVICE_TREE="sun8i-h2-plus-orangepi-zero"
++CONFIG_ENV_IS_IN_MMC=y
```
mirzak: Would like to avoid this one, because this will be needed for every single sunxi board. One example is that I have a Orange Pi PC Plus which no longer works unless I do the same changes as done here.

meta-mender-sunxi/recipes-bsp/u-boot/files/0002-env-Kconfig-remove-defaults-for-SUNXI.patch
nandra: Oh right, then it complain that ENV_OFFSET is empty and could be 0x800000 so should I drop it?

meta-mender-sunxi/recipes-bsp/u-boot/files/0003-Enable-env-is-in-mmc-define-offset.patch
nandra: Yes I should properly review patches ;). Thanks.

meta-mender-sunxi/recipes-bsp/u-boot/files/0003-Enable-env-is-in-mmc-define-offset.patch
nandra: Well yes but do you suggest to put it to sunxi-common.h ?

nandra:
With changes on https://github.com/mirzak/meta-mender-community-1/commits/thud-update it builds fine. Thanks.

mirzak:
Does it work as well? :D

nandra:
Sure :+1: thanks for support

mirzak:
Closing this, replaced by #36

mirzak:
And thank you for your work @nandra

PR #34, opened 2018-12-10
sunxi - thud update

nandra:
Tested on orangepi zero

PR #36, opened 2018-12-11
Thud update

mirzak:
Supersedes #35 and #34


mirzak:
Tested on Orange Pi PC Plus

PR #46, opened 2019-03-05
Rocko compulab

drewmoseley:


drewmoseley:
Note in particular https://github.com/mendersoftware/meta-mender-community/pull/46/commits/99ed313d5c7b0dc96ad6bf38a3e6dd92ad7f3b13 which will obviously affect other platforms.  I didn't see a way to make that change conditional to the FSLC based builds.

meta-mender-compulab/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Setup environment
+
+    $ export MACHINE=cl-som-imx8
+    $ source fsl-setup-release.sh -b build
+    $ source ../sources/meta-compulab-bsp/tools/setup-compulab-env
+    $ cat ../sources/meta-mender-community/meta-mender-compulab/templates/bblayers.conf.append >> conf/bblayers.conf
+    $ sed -i 's@^\(PREFERRED_RPROVIDER_u-boot-fw-utils.*$\)@# commented by Mender # \1@' conf/local.conf
+    $ sed -i 's@^\(CORE_IMAGE_EXTRA_INSTALL.*$\)@# commented by Mender # \1@' conf/local.conf
+    $ sed -i 's@^\(IMAGE_FEATURES.*package-management.*$\)@# commented by Mender # \1@' conf/local.conf
```
mirzak: I am curious what these `sed -i` commands are removing and why this can not be in the vendor specific  `local.conf.append`

scripts/mender.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
   <manifest>
-  <default sync-j="4" revision="rocko"/>
```
mirzak: I guess it is ok to remove this as we typically set the `revision` for each layer. We have done at least so far.

One note about the wording of the commit message, 

      Notably this will not coexist with the Freescale Community BSP

Compulab actually is using the NXP BSP, do not know if the problem would have been the same with the community one.

meta-mender-compulab/README.md
drewmoseley: The local.conf provided by the vendor setup-environment scripts includes imx examples, package-management and a number of other things that are beyond our scope here.

drewmoseley: Note that I am _not_ using the standard mender-community setup-environment here.

mirzak: > Note that I am not using the standard mender-community setup-environment here.

Aware of this, and that is why I am curios why you are modifying the functionality of the upstream BSP if the purpose is to integrate tightly with it? Defeats the purpose a bit, and might have as well done a "mender-community" environment in that case.

I know that I am nit picky here but just want to make sure that we agree/comply with some sort of policy.

drewmoseley: Yeah, understood. This is a bit of a conflict but we also generally only produce small demo images here (thus the inclusion of meta-mender-demo) so I thought in this case removing those bits was ok. I know as a _user_ of that BSP I would likely be removing them.  I can drop this if you think that's a better approach.

mirzak: In the "meta-mender-community" case then we are just building `core-image-base` together with meta-mender-demo which is just standard poky stuff. 

But as the purpose of integrating more tightly with a vendor BSP is to be as close as possible to the vendor BSP for this reason I think it is better to drop these.

drewmoseley: Hi @mirzak I updated per your suggestions.  Note that I specifically had to increase the TOTAL_SIZE_MB in this case and add https://github.com/mendersoftware/meta-mender-community/pull/46/files#diff-fa575c1f7060d27460b08cba85c7bb88R2 to overrule one of the defaults in the vendor provided local.conf file.

mirzak: Thanks, and the additional changes makes sense  since they are related to Mender working together with the BSP.

PR #55, opened 2019-04-09
Initial variscite support - var-som-solo

drewmoseley:


meta-mender-variscite/recipes-bsp/u-boot/u-boot-variscite-mender-common.inc
```diff
@@ -1,5 +1,16 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
 
-BOOTENV_SIZE_imx6ul-var-dart = "0x00002000"
+#### BOOTENV_SIZE_imx6ul-var-dart = "0x00002000"
+#### 
+#### SRC_URI_append_imx6ul-var-dart = " file://0001-CONFIGS-mx6_ul_var_dart-add-Mender-support.patch"
```
mirzak: Leftover commented code?

drewmoseley: I have a Dart here but have not had a chance to test it yet. I'll get to that soon but for now I figured I'd just leave the dart code here since it's not hurting anything.

mirzak: Agree that it is ok to keep the dart stuff around and this stuff has really not been tested before either. I just feel that there is no need to comment the above section, and instead just leave it as is.

drewmoseley:
@mirzak this should be ready to go

mirzak:
Nice

PR #58, opened 2019-04-19
Tinker updates

drewmoseley:
@mirzak I was playing a bit with the tinkerboard and needed these updates.  Are these generally useful?

### DIFF
--- a/meta-mender-rockchip/recipes-bsp/grub-mender-grubenv/files/02_tinkerboard_console_grub.cfg
+++ b/meta-mender-rockchip/recipes-bsp/grub-mender-grubenv/files/02_tinkerboard_console_grub.cfg
@@ -1,0 +1,1 @@
+set console_bootargs="console=ttyS2,115200n8"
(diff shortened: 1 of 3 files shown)


meta-mender-rockchip/recipes-bsp/grub-mender-grubenv/files/02_tinkerboard_console_grub.cfg
```diff
@@ -0,0 +1 @@
+set console_bootargs="console=ttyS2,115200n8"
```
mirzak: What is the reasoning behind this change?

I assume the default will output console on "serial + hdmi" during U-boot and in user-space, and kernel boot log is only printed on HDMI.

I am not sure if we should change this.

drewmoseley: In my case I had neither serial or HDMI console without this. I'll test again with a clean build but it looks like something is not right by default

drewmoseley:
I no longer have access to this hardware.  Closing.

PR #60, opened 2019-05-07
add meta-mender-update-modules layer with the first module

mirzak:
meta-mender-update-modules is the intended location where one can share Yocto recipes for specific Update Modules, in the simplest form it is simply a recipe installing the Update Module on the device. 

More advanced use-cases would be enabling creating of Update Module artifacts during an Yocto build.

### DIFF
--- a/meta-mender-update-modules/README.md
+++ b/meta-mender-update-modules/README.md
@@ -1,0 +1,43 @@
+meta-mender-update-modules
+==========================
+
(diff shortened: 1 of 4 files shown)


texierp:
Nice !
Just a small suggestion, maybe we can follow the standard described in [1] to reorganize the recipes ?

For instance, inherit allarch before do_install().

[1] https://www.openembedded.org/wiki/Styleguide

/Pierre-Jean

PR #67, opened 2019-05-14
Use official release of meta raspberry pi instead of the developers repository

sensslen:
This Patch uses the official meta raspberrypi instead of the developers


mirzak:
Hi @sensslen, thanks for taking the time to create this.

But https://github.com/agherzan/meta-raspberrypi is the official repository and the one hosted on Yocto is just a mirror of the first.

PR #71, opened 2019-05-21
A few more variscite changes.

drewmoseley:


drewmoseley:
This now supports both var-som-solo and imx6ul-var-dart.  SDCard only
For now I'm not going to backport emmc or NAND to this branch.
@mirzak please review for merge.

PR #75, opened 2019-06-03
NAND support for Variscite var-som-solo

drewmoseley:


drewmoseley:
And now, NAND on imx6ul-var-dart is validated.

drewmoseley:
I also cleaned up and tested in all configurations, the loading of the splash bitmap from the active root filesystem.

PR #77, opened 2019-06-25
eMMC/NAND support

drewmoseley:
This has been backported from thud for a customer deliverable.  Please review.


mirzak:
Most of the commits here are cherry-picks right? If they are it would have been to have that in the commit messages (`git cherry-pick -x`).

For future reference, it would be nice to keep an prefix on the commits as you have on some, e.g:

    variscite: <message>

Without this, it becomes very hard to read the git log to get an overview. And one needs to inspect each patch to figure out that it actually does not relate to whatever layer I am working on.

drewmoseley:
Updated as requested

PR #81, opened 2019-07-10
imx: move MENDER_STORAGE_DEVICE definitions to local.conf.bbappend

mirzak:
Having the definition in a IMAGE_CLASS will make these only applicable
on the image built, but this is not applied when you build e.g u-boot
which also needs a correct definition of MENDER_STORAGE_DEVICE.


mirzak:
@thochstein, can you review as well?

thochstein:
Looks good to me.

PR #82, opened 2019-07-17
mender: Add commented values for storage configuration.

drewmoseley:


drewmoseley:
This is based on user-feedback.  Some platforms rely on the defaults and it may be a bit confusing when the build fails due to the image being too large.

templates/local.conf.append
```diff
@@ -49,3 +49,23 @@ VIRTUAL-RUNTIME_initscripts = ""
 #MENDER_SERVER_URL = "https://docker.mender.io"
 #FILESEXTRAPATHS_prepend_pn-mender := "<DIRECTORY-CONTAINING-server.crt>:"
 #SRC_URI_append_pn-mender = " file://server.crt"
+
+# Mender storage configuration
+#
+# More details on these variables is available at
+#    https://docs.mender.io/2.0/devices/yocto-project/partition-configuration#configuring-storage
```
mirzak: Can you drop the `2.0` from the link.

drewmoseley: Done

PR #83, opened 2019-07-17
mender: Add commented values for storage configuration.

drewmoseley:


drewmoseley:
Remove version from docs link as well.

PR #93, opened 2019-08-29
Add clearfog-pro support.

Valks:
This layer depends upon pull request: [https://github.com/mirzak/meta-clearfog/pull/3](https://github.com/mirzak/meta-clearfog/pull/3)

Adds support for the clearfog-pro, main change compared to the clearfog branch is support for the 6 port switch.

### DIFF
--- a/meta-mender-clearfog/templates/local.conf.append
+++ b/meta-mender-clearfog/templates/local.conf.append
@@ -1,12 +1,21 @@
 # Appended fragment from meta-mender-community/meta-mender-clearfog/templates
 
+# Uncomment the lines below depending on the model
+
+# Supported Machine Types:
+# - clearfog-base
+# - clearfog-pro
 MACHINE ?= "clearfog-base"
+
+# Set the storage medium
+# Refer: https://github.com/mirzak/meta-clearfog/blob/master/conf/machine/include/clearfog-common.inc#L8-L18
+UBOOT_BINARY ?= "u-boot-spl-sdhc.kwb"
 
 IMAGE_INSTALL_append = " kernel-image kernel-devicetree"
 
+MENDER_IMAGE_BOOTLOADER_FILE = "u-boot-{MACHINE}.bin"
+MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET = "1"
+
 MENDER_FEATURES_ENABLE_append = " mender-uboot mender-image-sd"
 MENDER_FEATURES_DISABLE_append = " mender-grub mender-image-uefi"
-
-MENDER_IMAGE_BOOTLOADER_FILE = "u-boot-clearfog-base.bin"
-MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET = "1"
 
--- a/meta-mender-clearfog/recipes-bsp/clearfog-u-boot-scr/files/boot.cmd.in
+++ b/meta-mender-clearfog/recipes-bsp/clearfog-u-boot-scr/files/boot.cmd.in
@@ -2,8 +2,8 @@ setenv bootargs 'root=${mender_kernel_root} rw rootwait'
 run mender_setup;
 
 # Boot the board
-load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage
-load ${mender_uboot_root} ${fdt_addr_r} /boot/armada-388-clearfog-base.dtb
+load ${mender_uboot_root} ${kernel_addr_r} /boot/@@KERNEL_IMAGETYPE@@
+load ${mender_uboot_root} ${fdt_addr_r} /boot/@@KERNEL_DEVICETREE@@
 
 bootz ${kernel_addr_r} - ${fdt_addr_r}
 


Valks:
I agree it should work with minimal work (if any). That said I want to keep the changes to boot.cmd.in because it fixed the path issue when looking for the right kernel image.

Also probably worth adding a few comments into the local.conf.append to help explain how to set the machine and parameters to set for emmc support.

I'll take a look at it next week.

mirzak:
> I agree it should work with minimal work (if any). That said I want to keep the changes to boot.cmd.in because it fixed the path issue when looking for the right kernel image.

Yeah that makes sense.

> Also probably worth adding a few comments into the local.conf.append to help explain how to set the machine and parameters to set for emmc support.

Yes, and we typically accommodate these integrations with a post on hub.mender.io, eg.

https://hub.mender.io/t/solidrun-clearfog-base/370/11

And this should cover how to build for one specific board. But comments in e.g `local.conf.append` is a good compliment.

Valks:
pending meta-clearfog merging https://github.com/mirzak/meta-clearfog/pull/3 the changes are ready.

Note: Added comments to help configure the layer.

Valks:
Note: Signed the commits.

Valks:
Don't merge yet, there's an issue with eMMC devices (haven't checked sd yet) on boot with systemd failing, need to figure out the cause.

```
[   10.435086] sd 4:0:0:0: [sda] 60496896 512-byte logical blocks: (31.0 GB/28.8 GiB)
[   10.442798] sd 4:0:0:0: [sda] Write Protect is off
[   10.447716] sd 4:0:0:0: [sda] No Caching mode page found
[   10.453044] sd 4:0:0:0: [sda] Assuming drive cache: write through
[   10.463837]  sda: sda1 sda2
[   10.468484] sd 4:0:0:0: [sda] Attached SCSI removable disk
[   10.483025] EXT4-fs (mmcblk0p2): mounted filesystem with ordered data mode. Opts: (null)
[   10.491183] VFS: Mounted root (ext4 filesystem) on device 179:2.
[   10.499234] devtmpfs: mounted
[   10.502687] Freeing unused kernel memory: 1024K
[   10.546144] Run /sbin/init as init process
[   10.724541] systemd[1]: Failed to insert module 'autofs4': No such file or directory
[   10.733636] systemd[1]: Failed to mount tmpfs at /sys/fs/cgroup: No such file or directory
[   10.742812] systemd[1]: Failed to mount cgroup at /sys/fs/cgroup/systemd: No such file or directory
[!!!!!!] Failed to mount API filesystems, freezing.
[   10.786096] systemd[1]: Freezing execution.
[   73.746383] random: crng init done
```

Valks:
Well I'm a little stumped with this one.

Looking in the ```linux-stable/4.19.28-4.19``` directory I can see ```defconfig``` and ```systemd.cfg``` in the root folder.

Unfortunately it seems that systemd.cfg isn't been applied, when I manually added the lines to defconfig it worked. Any thoughts on why it isn't using the config fragment?

Valks:
The issue with systemd is related to [https://github.com/mirzak/meta-clearfog/](https://github.com/mirzak/meta-clearfog/) and not meta-mender-community.

When I wrote the issue I lost track of which layer I was looking at but it still is a blocking issue.

Valks:
Issue has been fixed, when concatenating the files it was missing a new line. Have updated the meta-clearfog layer with a fix.

meta-mender-clearfog/templates/local.conf.append
```diff
@@ -1,12 +1,32 @@
 # Appended fragment from meta-mender-community/meta-mender-clearfog/templates
 
+# Uncomment the lines below depending on the model
+
+# Clearfog - Base
 MACHINE ?= "clearfog-base"
+MENDER_IMAGE_BOOTLOADER_FILE = "u-boot-clearfog-base.bin"
```
mirzak: We can probably set this to:

     MENDER_IMAGE_BOOTLOADER_FILE = "u-boot-${MACHINE}.bin"

To avoid duplicate entries.

meta-mender-clearfog/templates/local.conf.append
```diff
@@ -1,12 +1,32 @@
...
+# UART (Serial)
+#UBOOT_BINARY ?= "u-boot-spl-uart.kwb"
 
 IMAGE_INSTALL_append = " kernel-image kernel-devicetree"
 
 MENDER_FEATURES_ENABLE_append = " mender-uboot mender-image-sd"
 MENDER_FEATURES_DISABLE_append = " mender-grub mender-image-uefi"
 
-MENDER_IMAGE_BOOTLOADER_FILE = "u-boot-clearfog-base.bin"
-MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET = "1"
```
mirzak: You removed `MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET`, this is required for it to put the `MENDER_IMAGE_BOOTLOADER_FILE` at the correct location.

Is it something else then `1` on Clearfog Pro?

meta-mender-clearfog/templates/local.conf.append
```diff
@@ -1,12 +1,32 @@
...
+# SPI Interface
+#UBOOT_BINARY ?= "u-boot-spl-spi.kwb"
+# NAND
+#UBOOT_BINARY ?= "u-boot-spl-nand.kwb"
+# SATA
+#UBOOT_BINARY ?= "u-boot-spl-sata.kwb"
+# eMMC
+#UBOOT_BINARY ?= "u-boot-spl-mmc.kwb"
+# UART (Serial)
+#UBOOT_BINARY ?= "u-boot-spl-uart.kwb"
```
mirzak: Replace this with a link to,

https://github.com/mirzak/meta-clearfog/blob/master/conf/machine/include/clearfog-common.inc#L8-L18

To avoid duplication of the same information.

meta-mender-clearfog/templates/local.conf.append
Valks: Didn't even click to do that, good idea.

meta-mender-clearfog/templates/local.conf.append
Valks: I'm not sure when or why I removed it but it didn't end up causing an issue so I forgot about it.

I have restored the setting and tested with eMMC and there's no problems.

meta-mender-clearfog/templates/local.conf.append
Valks: Done

Valks:
Thanks for all the help and feedback, hopefully it's now ready.

mirzak:
Awesome, thank you for working on this. Merging.

mirzak:
Would be great if you are able to create a completing post on [Mender Hub](https://hub.mender.io/c/board-integrations/yocto-project)

mirzak:
> Would be great if you are able to create a complementing post on Mender Hub

Ping

Valks:
Post created, based on clearfog base: https://hub.mender.io/t/solidrun-clearfog-pro/1148

Working on warrior support and using the official BSP layer (https://github.com/SolidRun/meta-clearfog), PR hopefully today.

As a side bonus using the official BSP support for the Clearfog-GTR should come for free but I don't have a board to verify.

PR #94, opened 2019-09-10
clearfog: fix missing $ in local.append template

Valks:
Missed the $ sign in the ```local.conf.append``` file.

A little surprised since I compiled it prior to saying it's ready, must have had a stale build at the time.

Anyway a quick fix.


mirzak:
Thanks

PR #101, opened 2019-09-26
Warrior cleanups

drewmoseley:

PR #103, opened 2019-09-27
raspberrypi: add README containing maintainer information

mirzak:
Per our discussion in #101, @texierp @drewmoseley , here is a suggested template for sub-layers and the information that they should contain (including maintainer)

### DIFF
--- a/meta-mender-raspberrypi/README.md
+++ b/meta-mender-raspberrypi/README.md
@@ -1,0 +1,61 @@
+# meta-mender-raspberrypi
+
+Mender integration layer for Raspberry Pi family of boards. Note that this layer
+only contains templates, and the actual integration patches live in
+[meta-mender/meta-mender-raspberrypi](https://github.com/mendersoftware/meta-mender/tree/master/meta-mender-raspberrypi).
+
+The supported and tested boards are:
+
+- [Raspberry Pi 4 Model B](https://hub.mender.io/t/raspberry-pi-4-model-b/889/2)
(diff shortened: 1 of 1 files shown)


mirzak:
@nandra @dwalkes @Valks, feel free to comment as you also have created layers in this layer.

nandra:
:+1:

mirzak:
Updated with a `Dependencies` section, to make the README follow common practices for Yocto/OE-core layers.

drewmoseley:
Looks ok to me.  The only comment I have is do we really want the quick start section?  That gives us two places with the same info which need to be manually kept in sync. Should we just link to the hub post instead?

meta-mender-raspberrypi/README.md
```diff
@@ -0,0 +1,61 @@
+# meta-mender-raspberrypi
+
+Mender integration layer for Raspberry Pi family of boards. Note that this layer
+only contains templates, and the actually integration patches live in
```
dwalkes: Minor nit - I think this should be and the *actual* integration pactches insetad of and the *actually* integration packages.

dwalkes:
@drewmoseley @mirzak +1 for one place to update.  I've had a conversation with @mirzak about this in the past.  My personal preference is to keep more detail with the code rather than the mender hub page to more cleanly handle the cases when documentation changes between releases, however I can see value in both approaches.

meta-mender-raspberrypi/README.md
mirzak: Thanks fixed.

PR #108, opened 2019-10-04
Update odroid manifest revision on warrior branch

nickbetteridge:
Attempting to build on warrior for odroid-c2 - although still complains:
```ERROR: Layer mender is not compatible with the core layer which only supports these series: thud (layer is compatible with warrior)```


mirzak:
There is PR already to update the warrior support,

https://github.com/mendersoftware/meta-mender-community/pull/104

PR #111, opened 2019-11-08
QA-111: Check signed commits in community repositories

oleorhagen:
Changelog: None

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>


oleorhagen:
@lluiscampos if this is good, I will open PR's to all the other branches also.

lluiscampos:
Good point about the other branches! Go ahead :)

PR #119, opened 2019-12-17
meta-mender-tegra: add support for jetson-nano development kit

Ecordonnier:
Jetson-nano development kit has a SD-card slot and no emmc. It does not support
nvidia redundant boot feature and therefore there is no dependency to the BUP Payload which
is used on jetson-tx2. Nvidia released a production module of the jetson-nano with 16GB emmc,
but this is not supported by this branch.

Even though there is no emmc, the sd-card is referenced as mmc device in u-boot, which is
why we use ENVIRONMENT_IS_IN_MMC.

Changelog: Title

Signed-off-by: Moritz Marquardt <moritz.marquardt@zeiss.com>
Signed-off-by: Etienne Cordonnier <etienne.cordonnier@taurob.com>

### DIFF
--- a/meta-mender-tegra/README.md
+++ b/meta-mender-tegra/README.md
@@ -5,6 +5,7 @@ Mender integration layer for NVIDIA Tegra hardware.
 The supported and tested boards are:
 
 - [Jetson TX2](https://hub.mender.io/t/nvidia-tegra-jetson-tx2/123)
+- [Jetson Nano](https://hub.mender.io/t/nvidia-tegra-jetson-nano/1360)
 
 Visit the individual board links above for more information on status of the
 integration and more detailed instructions on how to build and use images
@@ -43,7 +44,10 @@ repo init -u https://github.com/mendersoftware/meta-mender-community \
            -b warrior
 repo sync
 source setup-environment tegra
+# either:
 MACHINE=jetson-tx2 bitbake core-image-base
+# or:
+MACHINE=jetson-nano bitbake core-image-base
 ```
 
 
--- a/meta-mender-tegra/classes/image_types_mender_tegra.bbclass
+++ b/meta-mender-tegra/classes/image_types_mender_tegra.bbclass
@@ -12,3 +12,10 @@ tegraflash_create_flash_config_append() {
         -e"s,DATAFILE,${DATAFILE}," \
         flash.xml.in
 }
+
+tegraflash_create_flash_config_tegra210_append() {
+    sed -i \
+        -e"s,DATAFILE,${DATAFILE}," \
+        -e"s,APPSIZE,$(expr ${IMAGE_ROOTFS_SIZE} '*' 1024),g" \
+        $destdir/sdcard-layout
+}
--- a/meta-mender-tegra/classes/tegra-mender-setup.bbclass
+++ b/meta-mender-tegra/classes/tegra-mender-setup.bbclass
@@ -14,18 +14,22 @@ PREFERRED_RPROVIDER_u-boot-fw-utils = "u-boot-fw-utils-tegra"
 # Note: this isn't really a boot file, just put it here to keep the mender build from
 # complaining about empty IMAGE_BOOT_FILES.  We won't use the full image anyway, just the mender file
 IMAGE_BOOT_FILES = "u-boot-dtb.bin"
-# Mender customizations to support jetson tx2.  This needs to match up with flash_l4t_t186.custom.xml scheme
-# You will need to update these partition values when you update the flash layout.  One way to find the correct number is to
+# Mender customizations to support jetson tx2 and jetson nano.  This needs to match up with flash_l4t_t186.custom.xml/sdcard-layout-mender.in scheme
+# You will need to update these partition values when you update the flash/sd-card layout.  One way to find the correct number is to
 # boot into an emergency shell and examine the /dev/mmcblk* devices, or use the uboot console to look at mtdparts
-MENDER_DATA_PART_NUMBER = "31"
+MENDER_DATA_PART_NUMBER = "${@'15' if d.getVar('MACHINE') == 'jetson-nano' else '31'}"
 MENDER_ROOTFS_PART_A_NUMBER = "1"
-MENDER_ROOTFS_PART_B_NUMBER = "30"
+MENDER_ROOTFS_PART_B_NUMBER = "${@'14' if d.getVar('MACHINE') == 'jetson-nano' else '30'}"
 
 # Use a 4096 byte alignment for support of tegraflash scheme and default partition locations
 MENDER_PARTITION_ALIGNMENT = "4096"
 
-# Use no reserved space for bootloader data, since we will store in the partition block for the image
-MENDER_RESERVED_SPACE_BOOTLOADER_DATA = "0"
+# For jetson-tx2, use no reserved space for bootloader data, since we will store u-boot environment in the emmc boot partition and will use 0 bytes of the user
+# part of the emmc
+MENDER_RESERVED_SPACE_BOOTLOADER_DATA_tegra186 = "0"
+
+# For jetson-nano, u-boot environment gets stored in the first partition of the SD-card. Use 2 times u-boot's BOOTENV_SIZE (0x20000)
+MENDER_RESERVED_SPACE_BOOTLOADER_DATA_tegra210 = "262144"
 
 # See note in https://docs.mender.io/1.7/troubleshooting/running-yocto-project-image#i-moved-from-an-older-meta-mender-branch-to-the-thud-branch-and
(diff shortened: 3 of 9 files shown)


mirzak:
Nice work! This looks good to me. @dwalkes can I get an ack from you as well before merging.

Ecordonnier:
Thanks! I fixed a comment and repushed the branch ("u-boot environment gets stored in the first partition" -> "u-boot environment gets stored in the ENV partition").

dwalkes:
> @dwalkes can I get an ack from you as well before merging.

All open issues at https://github.com/Trellis-Logic/meta-mender-community/pull/9 are resolved and I see no issues with code review but if I'm going to maintain this platform I would prefer to try the build on hardware before merging.  I thought I'd be able to try today but I wasn't able to get hardware from CU as I had hoped.  I ordered a board today, should be able to try it next week.

Also based on the comments in https://github.com/Trellis-Logic/meta-mender-community/pull/7 I don't think anyone has tried rollback or the integration list yet, and I'd like to do that for completeness.

I'll check back in when I get hardware, hopefully next week.

Ecordonnier:
I had tested manual rollback (with "mender -rollback") and it worked.

Those are my checklist results as reference (I am not using mender as systemd service):
[checklist-6.txt](https://github.com/mendersoftware/meta-mender-community/files/3982646/checklist-6.txt)
[checklist-9.txt](https://github.com/mendersoftware/meta-mender-community/files/3982647/checklist-9.txt)
[checklist-18.txt](https://github.com/mendersoftware/meta-mender-community/files/3982648/checklist-18.txt)

dwalkes:
I have my shiny new nano board in hand and tried this out today.

One thing I noticed is `doflash.sh` as used on the TX2 and as documented in https://github.com/madisongh/meta-tegra/wiki/Flashing-the-Jetson-Dev-Kit doesn't work for me.  It looks like there are at least two issues, one is that doflash.sh references the flash.xml.in instead of sdcard.xml.in, looks like this is [hard coded here](https://github.com/madisongh/meta-tegra/blob/warrior/classes/image_types_tegra.bbclass#L395).  Second issue is that even with this change boot halts during u-boot, with this message
```
MMC:   Tegra SD/MMC: 0, Tegra SD/MMC: 1
In:    serial
Out:   serial
Err:   serial
Net:   No ethernet found.
Failed to mount ext2 filesystem...
** Unrecognized filesystem type **
Hit any key to stop autoboot:  0
Retrieving file: /boot/extlinux/extlinux.conf
Failed to mount ext2 filesystem...
** Unrecognized filesystem type **
Error reading config file
Tegra210 (P3450-Porg) #
```
I haven't dug in much further to this one.  It looks like the partition table is there on mmc 1, not sure why the filesystem isn't recognized.  I thought it might be related to https://github.com/Trellis-Logic/meta-mender-community/pull/8 but cherry picking that commit didn't help.

I then noticed @Ecordonnier's page [on mender hub](https://hub.mender.io/t/nvidia-tegra-jetson-nano/1360) which uses `dd` on the build system instead of tegraflash.  That's a simpler option for Nano and the best way to support without needing to keep up with tegraflash inevitable headaches.  I've confirmed this works for me.  So I'm not sure we care if `deploy.sh` and tegraflash doesn't work on the Nano dev board with sdcard.  However I think I'll open an issue to track on meta-tegra and update the wiki page, assuming you are also not successful or haven't tried using `doflash.sh` and tegraflash.

The [integration checklist](https://docs.mender.io/2.2/devices/yocto-project/bootloader-support/u-boot/integration-checklist) completes with no issues for me including u-boot environment manipulation from the u-boot prompt.

I did notice another set of patch fuzz on the latest warrior branch commit, which I've fixed in https://github.com/taurob/meta-mender-community/pull/1 , fine to squash that one into yours and force push if you prefer @Ecordonnier

Great work @Ecordonnier and @moqmar and thanks for your contribution!

mirzak:
@dwalkes @Ecordonnier @moqmar, is this good to go?

dwalkes:
@mirzak 

There's one more commit which @Ecordonnier could merge in [here](https://github.com/taurob/meta-mender-community/pull/1), for patch fuzz, might be worth waiting until he's back next week to merge those but if you are in a hurry to close I think it's fine to merge as is.

Note: there are also some pending related changes for zeus on https://github.com/Trellis-Logic/meta-mender-community/pull/10 but I think it's probably worthwile to get the warrior support for Nano in as-is, unless @Ecordonnier plans to backport https://github.com/Trellis-Logic/meta-mender-community/pull/10 to warrior or abandon this branch in favor of the zeus implementation in https://github.com/Trellis-Logic/meta-mender-community/pull/10.  @mirzak when do you anticipate we'll have a zeus branch ready on meta-mender-community?

Ecordonnier:
@mirzak  Technically, I think this branch would be ready to be merged.

However you should really consider the long-term strategy regarding the merge of https://github.com/Trellis-Logic/meta-mender-community/pull/10, because it may be easier to drop my branch entirely and merge only madisongh's branch, rather than merging my branch into warrior and then madisongh's branch into zeus, which may make mender updates from the warrior branch to the zeus branch on Jetson-Nano extremely complicated, since the two branches are not implementing the update in the same way (the rootfs size is different for instance, and u-boot environment is in the SD-card in my branch but in the SPI flash in madisongh's branch).

Also you need to consider that madisongh is a long-term maintainer of meta-tegra, whereas I can work on this only for a few weeks, plus the fact that madisongh's branch is adding support for all meta-tegra platforms and not only for jetson-nano, so I think the best long-term option is to go for his branch once it is ready.

I would suggest that https://hub.mender.io/t/nvidia-tegra-jetson-nano/1360 gets reused for madisongh's branch and to just change the link to the pull request.

dwalkes:
@Ecordonnier 

> Also you need to consider that madisongh is a long-term maintainer of meta-tegra, whereas I can work on this only for a few weeks, plus the fact that madisongh's branch is adding support for all meta-tegra platforms and not only for jetson-nano, so I think the best long-term option is to go for his branch once it is ready.

I absolutely agree with this statement.

> which may make mender updates from the warrior branch to the zeus branch on Jetson-Nano extremely complicated, since the two branches are not implementing the update in the same way (the rootfs size is different for instance, and u-boot environment is in the SD-card in my branch but in the SPI flash in madisongh's branch).

I think updates across L4T releases/yocto branches for Nano are unlikely to work in either case, since [Redundant Boot Support is Different](https://docs.nvidia.com/jetson/archives/l4t-archived/l4t-3231/index.html#page/Tegra%2520Linux%2520Driver%2520Package%2520Development%2520Guide%2Fbootloader_update.html%23) and [not supported with the BUP scheme](https://github.com/Trellis-Logic/meta-mender-community/pull/10/commits/82ae4047525e8fccc485f901d48e538f36ebf742) in https://github.com/Trellis-Logic/meta-mender-community/pull/10.

If you don't plan to add SD-card based boot to https://github.com/Trellis-Logic/meta-mender-community/pull/10 and won't need it going forward for your project on zeus and later I think I would agree we should leave it out of meta-mender-community and just going with the implementation in https://github.com/Trellis-Logic/meta-mender-community/pull/10 as-is until someone has a need for it again on later branches.  I haven't looked at the complexity of adding it myself.

So in this case I suggest we change the mender hub page for  to reference this PR for the warrior branch but don't merge the branch, list zeus as a work in progress and link to https://github.com/Trellis-Logic/meta-mender-community/pull/10.

@mirzak what do you think?

dwalkes:
> If you don't plan to add SD-card based boot to Trellis-Logic#10 and won't need it going forward for your project on zeus and later I think I would agree we should leave it out of meta-mender-community and just going with the implementation in Trellis-Logic#10 as-is until someone has a need for it again on later branches. I haven't looked at the complexity of adding it myself.

@madisongh has added this option in the PR at #10 as well, see https://github.com/Trellis-Logic/meta-mender-community/pull/10/commits/97139c6f1a9578f60306650d85ebeaa5533d9f44.  So in my opinion we might as well merge the changes in this PR just because it will make warrior based installs easier for anyone trying to use the nano until meta-mender and meta-mender-community zeus support is ready.    Where there are conflicts with the changes in this branch on the merge to zeus I will plan on using the changes in zeus and Trellis-Logic#10.

mirzak:
I apologize for the delay from my, have been out of office and traveling the past few weeks and has been difficult to stay on top of things :).

This is awesome work done from everyone invalided.

> So in my opinion we might as well merge the changes in this PR 

I will do this then.

In regards to `zeus` support, we are probably looking at another 4 weeks before it is added to meta-mender, though I would open to create a `zeus` branch ahead of that in meta-mender-community to get https://github.com/Trellis-Logic/meta-mender-community/pull/10 in if it is in a mergable state.We did something similar for the warrior branch.

PR #130, opened 2020-03-24
Rocko backports

drewmoseley:
Needed on this branch for a customer project.


mirzak:
Can you do thud and sumo while you are at it?

PR #134, opened 2020-03-29
initial support for Google Coral Dev Board

mirzak:

### DIFF
--- a/meta-mender-coral/recipes-bsp/coral-boot-script/cora-boot-script.bbappend
+++ b/meta-mender-coral/recipes-bsp/coral-boot-script/cora-boot-script.bbappend
@@ -1,0 +1,1 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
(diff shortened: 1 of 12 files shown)


meta-mender-coral/recipes-bsp/coral-boot-script/cora-boot-script.bbappend
```diff
@@ -0,0 +1 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
```
drewmoseley: Is this file name right? It's missing the "l".

mirzak: It is actually correct, the upstream recipe has a typo. I will blame the upstream maintainer on this one :)

I did create a [PR](https://github.com/mirzak/meta-coral/pull/15) to fix this in master in upstream, but I do not know if I dare to backport this now.

PR #165, opened 2020-08-10
meta-mender-tegra: Fix delta updates

nielsavonds:
Mount old and new filesystem as read-only to avoid breaking delta-updates

Signed-off-by: Niels Avonds <niels@nobi.life>

### DIFF
--- a/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-reboot-state-script
+++ b/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-reboot-state-script
@@ -9,17 +9,22 @@ if [ $num_slots != 2 ]; then
 fi
 
 new_boot_part=`fw_printenv -n mender_boot_part`
+
 new_mender_boot_part_mountpoint=/tmp/new_mender_boot_part
+ota_package_location=/opt/ota_package/bl_update_payload_current
+
 echo "Mounting new install partition ${new_boot_part}" >&2
 mkdir -p ${new_mender_boot_part_mountpoint}
-mount /dev/mmcblk0p${new_boot_part} ${new_mender_boot_part_mountpoint}
-mount -oremount,rw /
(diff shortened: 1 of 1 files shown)


mirzak:
Looping in @dwalkes

nielsavonds:
A similar fix is needed in the Zeus branch. I think it would be simpler there, since bind mounts are already being used. It's probably as simple as adding "-o ro" to the mount in the Zeus branch.

dwalkes:
Thanks very much @nielsavonds this looks good to me.  We will make sure this is incorporated on the dunfell branch we are currently working in on https://github.com/OE4T/meta-mender-community/pull/1 and get it backported to Zeus from there as well.  I actually haven't tested delta updates myself yet but this is something we need to try.

kekiefer:
Note that since f0dcd9100566fe0ee0833ebb89c919beefb62cb3 from zeus, a `chroot` is used to access the new update payload instead of copying it back.

nielsavonds:
I don't know exactly what changes, but mounting a filesystem as read-write just once changes the checksum of that filesystem and thus breaks delta updates. I tested this on my host and on the target and got similar results. Since we don't expect to write anything to the new partition anyway, a read-only mount seems safer.

kekiefer:
At least in the newer `nv_update_engine`, it creates an `entry_table` file in `/opt/ota_package` https://github.com/OE4T/meta-mender-community/pull/1#issuecomment-671440073 and offhand I'm not sure if it absolutely requires this, or will ignore errors here. So it remains to be seen if we can do this in newer L4T releases or we have to get a little fancier.

mirzak:
Is this good to go @dwalkes ?

dwalkes:
Yes, this is good to go into the the warrior branch based on @neilsavonds testing. We still need to figure out what if anything we need to do on later branches.

mirzak:
Thanks, merged.

PR #167, opened 2020-08-19
Zeus fixes for sunxi platform

nandra:
In this PR are fixes which was verified on orange-pi-zero and orange-pi-pc-plus. Both board boots fine.


mirzak:
Awesome. I think this looks good.

One small thing, please update the README.md in meta-mender-sunxi to reflect the zeus bump.

nandra:
@mirzak updated. When merged I'll update topics for sunxi to be compatible with zeus. Thanks.

mirzak (approved):

PR #169, opened 2020-08-27
Bump poky revisions to Dunfell (qemu, bbb,rpi,odroid)

oleorhagen:

PR #170, opened 2020-09-07
Bump: manifest-qemu poky revision to Dunfell

oleorhagen:
This is the first commit from #169 , built, and tested locally, and confirmed that it works.


mirzak:
Thanks, please also update https://github.com/mendersoftware/meta-mender-community/blob/dunfell/meta-mender-qemu/README.md

And update https://hub.mender.io/t/qemu-the-fast-processor-emulator/420/2 when this is merged :)

PR #177, opened 2020-10-13
Update Jetson Nano Mender environment in MMC patch

arnstein:
OE4T/meta-tegra#437 updated SRCREV for meta-tegra-uboot, and after that this patch didn't apply.
This one should work.
DISCLAIMER: I do not have a Nano to test this on.
Also, with OE4T/meta-tegra#438 we might see
a similar problem on the zeus-l4t-r32.3.1 branch. I am unfortunately not able to test this.

Signed-off-by: Arnstein Kleven <arnsteinkleven@gmail.com>

### DIFF
--- a/meta-mender-tegra/recipes-bsp/u-boot/patches/0013-Jetson-Nano-mender-environment-in-mmc.patch
+++ b/meta-mender-tegra/recipes-bsp/u-boot/patches/0013-Jetson-Nano-mender-environment-in-mmc.patch
@@ -1,24 +1,13 @@
-From 0972c8b7ecc95ac27ef69bd4441154f422413bfd Mon Sep 17 00:00:00 2001
-From: Moritz Marquardt <moritz.marquardt@zeiss.com>
-Date: Thu, 5 Dec 2019 14:26:57 +0000
-Subject: [PATCH] Configure u-boot env. in sd-card for jetson-nano
-
----
- include/configs/p3450-porg.h | 4 ++--
- 1 file changed, 2 insertions(+), 2 deletions(-)
-
 diff --git a/include/configs/p3450-porg.h b/include/configs/p3450-porg.h
-index e8a7ead5bb..b818aae159 100644
+index 7be05f3dae..2e97052610 100644
 --- a/include/configs/p3450-porg.h
 +++ b/include/configs/p3450-porg.h
-@@ -34,8 +34,8 @@
- 	func(PXE, pxe, na) \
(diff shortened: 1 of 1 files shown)


mirzak:
@dwalkes, can you have a look?

dwalkes:
@arnstein thanks for contributing

@mirzak I haven't had a chance to test this yet but it looks very low risk to me.  If you'd like me to test I'll try to get to it this weekend.

dwalkes:
> The ENV offset is using the CONFIG_SPI_FLASH_SIZE here. Maybe this is overridden later on somewhere, but I think this might be worth to test.

Yea good point, I think it's set here:
https://github.com/mendersoftware/meta-mender-community/blob/a828ac133c1295c560d26cc50d20c2c83e742c0a/meta-mender-tegra/recipes-bsp/u-boot/patches/0012-Update-environment-defaults-for-tegra.patch#L40
but I'm not sure how we aren't ending up with a conflict.

The change in https://github.com/OE4T/u-boot-tegra/commit/e07c4410083f8c4439c5d00dc60a17488d07875b caused the patch conflict.

I'm not completely sure why we need to set any config environment at all in `include/configs/p3450-porg.h`, it seems like we'd just want to use `include/config_defaults.h`.

> Also, with OE4T/meta-tegra#438 we might see a similar problem on the zeus-l4t-r32.3.1 branch

I think we should be fine there since the only change to the u-boot repo was https://github.com/OE4T/u-boot-tegra/commit/a8c48a2aadb115ef9eb01fa5ab496a24fce65a25 which won't conflict with anything in configs, also this particular patch is not on [the zeus branch](https://github.com/mendersoftware/meta-mender-community/tree/zeus/meta-mender-tegra/recipes-bsp/u-boot/patches)

I'm tempted to just sunset the warrior branch support for Nano since we only supported MMC flash based boot there and we completely overhauled support in zeus and don't support that anymore as of dunfell.  @Ecordonnier or @moqmar as original contributors to this branch do you have any use for it going forward?

Ecordonnier:
Hi Dwalkes, I am not working with this hardware any more.

dwalkes:
Thanks @Ecordonnier 
@mirzak I'm not going to be able to get to this today but I think we should just keep this conversation history, not try to update the warrior branch, and close this merge request.  If someone really needs nano support on warrior it should be a solvable problem, but my first recommendation would be to move to dunfell or later branches since upgrades from warrior to later branches on nano isn't going to be easily supported.

PR #187, opened 2020-11-06
imx: fix missing "do_mender_tar_src" task

clementperon:
Hi @drewmoseley ,

I'm really new on Mender architecture.

I tried to add a new board using meta-mender-imx layer for testing and I got this the following error:
"do_mender_tar_src" is missing.

it looks like the u-boot-imx_% is missing u-boot-mender.inc

Could you review it, thanks for your help :)

### DIFF
--- a/meta-mender-imx/recipes-bsp/u-boot-imx/u-boot-imx_%.bbappend
+++ b/meta-mender-imx/recipes-bsp/u-boot-imx/u-boot-imx_%.bbappend
@@ -1,6 +1,10 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
+include ${@mender_feature_is_enabled("mender-uboot","u-boot-mender.inc","",d)}
+
 RPROVIDES_${PN}_mender-grub += "u-boot"
 RPROVIDES_${PN}_mender-uboot += "u-boot"
 
 SRC_URI_append_imx8mnevk = " file://0001-Switch-to-CONFIG_DISTRO_DEFAULTS-for-bootcmd.patch "
+
+MENDER_UBOOT_AUTO_CONFIGURE = "0"


mirzak:
Hmm, not sure this is needed.

My understanding is that the default configuration for all boards in `meta-mender-imx` is using the U-Boot/GRUB configuration, in which case you do not integrate directly with U-Boot so it should not be executing `do_mender_tar_src`.

Do you change the configuration in e.g `local.conf` ?  e.g do you have something like this:

```
MENDER_FEATURES_ENABLE_append = " mender-uboot mender-image-sd"
MENDER_FEATURES_DISABLE_append = " mender-grub mender-image-uefi"
```

If you do, you only need to apply it for the board you are trying to integrate instead of applying it for all.

clementperon:
@mirzak :
Yes indeed, the board I'm integrating use the u-boot-imx bootloader without grub as 2nd stage...

Using GRUB as 2nd stage is only required to support UEFI booting ?

mirzak:
> If this doesn't break the UBoot/GRUB config can this be apply ?

Yes, indeed. And looking at it again, it should not break the U-Boot/GRUB config but might need to test to verify.

meta-mender-imx/recipes-bsp/u-boot-imx/u-boot-imx_%.bbappend
```diff
@@ -1,6 +1,10 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
+require recipes-bsp/u-boot/u-boot-mender.inc
```
drewmoseley, line 3: Can you use this syntax instead?

`include ${@mender_feature_is_enabled("mender-uboot","u-boot-mender.inc","",d)}`

That matches with what is in master/dunfell better

drewmoseley (changes requested):
This seems ok with the one change I recommended.

meta-mender-imx/recipes-bsp/u-boot-imx/u-boot-imx_%.bbappend
clementperon: @drewmoseley done

drewmoseley (approved):

clementperon:
hi @drewmoseley @mirzak 

Sorry I didn't test this patch properly I was on zeus branch instead of patch-1...

Now the error is back. It's look like mender_feature_is_enabled function in my build.

This my machine-conf-mender.inc that I include 
`INHERIT += "mender-full"
IMAGE_CLASSES += "mender-setup-imx"

# Force to use u-boot-imx for easy intregation with meta-community-imx 
PREFERRED_PROVIDER_virtual/bootloader_mx8 = "u-boot-imx"

# Specify BOOTSIZE and ENV_OFFSET
BOOTENV_SIZE_imx8qm-board = "0x2000"
MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_1_imx8qm-board = "0x400000"
MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_2_imx8qm-board = "0x800000"

# Specify RootFS Storage on eMMC 0
MENDER_STORAGE_DEVICE_imx8qm-iwg27m = "/dev/mmcblk0"
MENDER_STORAGE_TOTAL_SIZE_MB_imx8qm-board = "2048"
MENDER_UBOOT_STORAGE_DEVICE_imx8qm-board = "0"

# Specify the bootloader file
MENDER_IMAGE_BOOTLOADER_FILE_imx8qm-board = "imx-boot"
IMAGE_BOOT_FILES_remove_imx8qm-board = " ${KERNEL_IMAGETYPE} ${@make_dtb_boot_files(d)} "

# Use mender with U-boot without GRUB nor UEFI support
MENDER_FEATURES_ENABLE_append = " mender-uboot mender-image-sd"
MENDER_FEATURES_DISABLE_append = " mender-grub mender-image-uefi"
`

clementperon:
Okay I found the issue

PR #188, opened 2020-11-08
meta-mender-tegra: dunfell nano development board partition layout fixes

dwalkes:
Per discussion at https://github.com/OE4T/meta-mender-community/pull/5

As of r32.3.1 (or possibly earlier) NVIDIA has defined a UBENV partition at offset 0x3D8000 for `jetson-nano-qspi-sd` partition layout, which is the layout used for the Nano development board with SD card.  Previously this area was defined as a `PAD` partition and we had stored our uboot environment in that location.  However, we'd also accidentally overflowed into the `VER_b` location with our definition of uboot storage environment from 0x3B0000 to 0x3F0000.  This might be a problem if/when we add support for [Nano redundant boot](https://docs.nvidia.com/jetson/l4t/index.html#page/Tegra%20Linux%20Driver%20Package%20Development%20Guide/bootloader_update_nano_tx1.html) for bootloader components.

The area defined by NVIDIA for `UBENV` is too small to store dual redundant copies of mender variables, so we need to stretch our area back into what was known as the `PAD` partition on previous releases and the unused space between `NXC_R` and `VER_b` partitions.

After some discussion in the previously mentioned thread, we've settled on keeping the `UBENV` starting at `0x3B0000` and reducing the size to 131072 bytes instead of 262144 of so we could continue to make the first uboot environment boundary at the same start location and complete both copies of uboot environment before the `VER_b` sector starts.  This means the u-boot environment for existing sdcard based Nano builds on the dunfell branch will move on the next tegraflash or `dosdcard.sh`  based update.  In the cvase of a `dosdcard.sh` based update the existing SPI flash u-boot environment will be invalid and replaced with defaults.

This alignment will be shared with the Jetson Nano 2GB hardware which will be brought in during gatesgarth/L4T+next (unless someone volunteers to add and backport it sooner in https://github.com/OE4T/meta-mender-community/issues/4) and the patches here were rebased from those in the above linked PR.

I've started a sheet at [this link](https://docs.google.com/spreadsheets/d/14zVltS2LTwCNywoje6_8wzFzHQ3sYVhY6C7T0DkOnx0/edit?usp=sharing) which I plan to try to update with each new release and platform to make sure we don't miss these layout changes going forward with new L4T releases.

I've completed the uboot integration tests at [this link](https://github.com/OE4T/tegra-demo-distro/wiki/Mender-Integration-Tests) for `jetson-nano-qspi-sd` with these changes.

### DIFF
--- a/meta-mender-tegra/recipes-bsp/tegra-binaries/mender-custom-flash-layout/jetson-nano-qspi-sd/flash_mender.xml
+++ b/meta-mender-tegra/recipes-bsp/tegra-binaries/mender-custom-flash-layout/jetson-nano-qspi-sd/flash_mender.xml
@@ -48,21 +48,22 @@
               binary. </description>
         </partition>
 
-        <!-- This is padding to ensure VER is at the end of flash -->
-        <partition name="PAD" id="6" type="data">
+        <partition name="UBENV" type="data">
             <allocation_policy> sequential </allocation_policy>
             <filesystem_type> basic </filesystem_type>
-            <size> 3342336 </size>
+	    <start_location> 0x3B0000 </start_location>
+            <size> 131072 </size>
             <file_system_attribute> 0 </file_system_attribute>
             <allocation_attribute> 8 </allocation_attribute>
             <percent_reserved> 0 </percent_reserved>
-            <description> Empty padding. </description>
+            <description> U-Boot environment area  </description>
         </partition>
 
-        <partition name="VER_b" id="7" type="data">
+        <partition name="VER_b" type="data">
             <allocation_policy> sequential </allocation_policy>
             <filesystem_type> basic </filesystem_type>
-            <size> 65536 </size>
+	    <start_location> 0x3E0000 </start_location>
+            <size> 32768 </size>
             <file_system_attribute> 0 </file_system_attribute>
             <partition_attribute> 0 </partition_attribute>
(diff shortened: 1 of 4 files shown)


mirzak:
> This probably means the u-boot environment for existing sdcard based Nano builds on the dunfell branch need to be re-tegraflashed, which hopefully isn't an issue since these devices aren't really meant for production anyway

What will happen if you do not do this and just deploy an update with a Mender Artifact (`.mender`) with these changes included? I assume this depends if U-Boot updates are enabled or not in the Nano board? If they are not enabled, there are no implications on existing devices I guess.

dwalkes:
> What will happen if you do not do this and just deploy an update with a Mender Artifact (.mender) with these changes included? I assume this depends if U-Boot updates are enabled or not in the Nano board? If they are not enabled, there are no implications on existing devices I guess.

Good point @mirzak.  Since we do not yet support bootloader updates on Nano I don't think anyone just performing mender updates will notice this change.  I think the only way they might be able to get confused would be if they updated the SDCard using the `dosdcard.sh` script, which would update both the bootloader and root filesystem but leave the u-boot environment alone in the SPI flash.  In this case I'd expect that on the next boot the old SPI flash environment would be corrupt (due to wrong size) and the boot would happen based on the default settings of boot partition, which is fine.

I've updated the description.

PR #191, opened 2020-11-13
Bsp v5.4.24

drewmoseley:
This fixes up the imx8mn-var-som machine so that it works with the 5.4.24 BSP from Variscite.
It also enabled mender-uboot integration in addition to the standard mender-grub integration.


mirzak (approved):

mirzak:
@drewmoseley , will leave to you to merge this if it is done

drewmoseley:
It's done for the one platform. I've not had a chance to check the others but they were on 5.4.3 last I checked.  I'll merge this one and then take another pass through with the boards that I have here.

PR #190, opened 2020-11-10
Fix patch so that it build with the latest NXP BSP

gphillips-a2e:
Submitting this as requested in https://hub.mender.io/t/variscite-var-som-mx8m-nano-nxp-i-mx-8m-nano/1928/3.
I am not confident of the changes as I don't know the code, and I am experiencing problems when running the image it generates.

PR #192, opened 2020-11-16
Support variscite dunfell branch

liuming50:


mirzak:
@drewmoseley , would you be able to give this a quick test?

drewmoseley:
I don't see anything wrong with this but it likely will still have the issues others have noted when using the Grub based integration (ie the kernel error messages reported when booting fsl-image-qt; [see here](https://github.com/mendersoftware/meta-mender-community/pull/190) for more details.

drewmoseley:
Should we wait until we get that fixed properly in Zeus and then move everything forward to Dunfell?

PR #206, opened 2020-12-14
meta-mender-tegra fixes for unexpected boot slot changes

madisongh:
The NVIDIA bootloaders implement a sort of boot counter (called "priority" in their tools), not mentioned in their documentation, that we need to account for.  The priority value needs to be reset on each successful boot to prevent the bootloaders from switching boot slots.

These patches add a service to update the priority value for the current boot slot on each successful boot.  It also removes some Tegra-specific state scripts that are obsoleted by the inclusion of this service.


dwalkes:
Thanks @madisongh 

@mirzak we've verified today with jetson-tx2 can xavier-nx cboot setups and some scripts at https://github.com/BoulderAI/meta-mender-community/commits/dunfell%2Bupdatebootloader%2Btests which I'll include in a follow-on PR.  This one should be ready to merge.

mirzak:
Thanks!

PR #207, opened 2020-12-15
Initial test scripts for mender on tegra

dwalkes:
Add a start on test scripts we can use to help validate some of the more complicated parts of mender integration on tegra, bootloader updates and boot slot matching.

This change has a working implementation for cboot based targets, and some work in progress for uboot based targets.  Uboot based targets need more work to support test automation and reliable bootloader updates, see discussions at https://github.com/BoulderAI/meta-mender-community/pull/1 and https://github.com/OE4T/meta-mender-community/pull/8#issuecomment-744376127

Signed-off-by: Dan Walkes <danwalkes@boulderai.com>


mirzak:
Thanks, and apologies for the delay.

PR #210, opened 2021-01-07
tegra: Use read only mounts for uboot and cboot builds

dwalkes:
This PR replaces https://github.com/mendersoftware/meta-mender-community/pull/208 and incorporates changes from @madisongh in https://github.com/OE4T/tegra-demo-distro/pull/31 to support read only rootfs mounts on both uboot and cboot tegra configurations during mender update, including persistent machine-id across updates in cboot.

Mender delta updates have not been specifically tested on cboot or uboot configurations with this PR but should be supported.


madisongh:
LGTM

mirzak:
Thank you

PR #223, opened 2021-03-31
Honor mender dtb variable

mrpelotazo:
Variable fdtfile ends up with a wrong value when using a custom device tree file (set via KERNEL_DEVICETREE). 

mender_dtb_name variable is set correctly though. So setting fdtfile to mender_dtb_name value fixes the issue.

Also on the same go, update my email address.

### DIFF
--- a/meta-mender-toradex-nxp/README.md
+++ b/meta-mender-toradex-nxp/README.md
@@ -79,6 +79,6 @@ The authors and maintainers of this layer are:
 
 - Mirza Krak - <mirza.krak@northern.tech> - [mirzak](https://github.com/mirzak)
 - Ricardo Sanchez - <ricardo.sanchez@aerin.es>
-- Adrian Antonana - <mrpelotazo@posteo.net>
+- Adrian Antonana - <adrian.antonana@plating.de>
 
 Always include the maintainers when suggesting code changes to this layer.
--- a/meta-mender-toradex-nxp/recipes-bsp/u-boot/files/colibri-imx6ull/0001-colibri-imx6ull-mender-manual-U-Boot-integration.patch
+++ b/meta-mender-toradex-nxp/recipes-bsp/u-boot/files/colibri-imx6ull/0001-colibri-imx6ull-mender-manual-U-Boot-integration.patch
@@ -1,7 +1,26 @@
-Index: git/configs/colibri-imx6ull_defconfig
-===================================================================
---- git.orig/configs/colibri-imx6ull_defconfig
-+++ git/configs/colibri-imx6ull_defconfig
+From ddaa62a9aa9ea2c63c9c3789717501438b6cc6a3 Mon Sep 17 00:00:00 2001
+From: Adrian Antonana <adrian.antonana@plating.de>
+Date: Wed, 31 Mar 2021 10:21:42 +0200
+Subject: [PATCH] colibri imx6ull mender manual U-Boot integration
+
+---
+ configs/colibri-imx6ull_defconfig | 10 +++++++---
+ include/configs/colibri-imx6ull.h | 17 ++++++++++-------
+ 2 files changed, 17 insertions(+), 10 deletions(-)
+
+diff --git a/configs/colibri-imx6ull_defconfig b/configs/colibri-imx6ull_defconfig
+index cf85716dae..30eac0bcbb 100644
+--- a/configs/colibri-imx6ull_defconfig
++++ b/configs/colibri-imx6ull_defconfig
+@@ -13,7 +13,7 @@ CONFIG_BOOTDELAY=1
+ # CONFIG_USE_BOOTCOMMAND is not set
+ # CONFIG_CONSOLE_MUX is not set
+ CONFIG_USE_PREBOOT=y
+-CONFIG_PREBOOT="setenv fdtfile imx6ull-colibri${variant}-${fdt_board}.dtb"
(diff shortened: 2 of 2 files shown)


mrpelotazo:
How is this looking?

drewmoseley:
I'm concerned that this will break other things.  I know that for many integrations, I use the `MENDER_UBOOT_PRE_SETUP_COMMANDS` variable to modify these types of values.  Typically I will use the same logic as provided by the BSP provider. In the case of Toradex, it's dynamically determined based on which carrier is installed.

Is there a reason you need to hard code it during the build?

mrpelotazo:
Hey, thanks for the reply!

To answer your question, here is how I end up doing it this way:

1. Independently of **KERNEL_DEVICETREE** containing multiple dtb files, mender considers only one:
```
meta-mender-core/classes/mender-helpers.bbclass
200:        bbwarn "Found more than one dtb specified in KERNEL_DEVICETREE (${KERNEL_DEVICETREE}). Only one should be specified. Choosing the last one: ${MENDER_DTB_NAME}. Set KERNEL_DEVICETREE to the desired dtb file to silence this warning."
```

To my understanding, even if the board is determined on runtime, there will be only one dtb built and installed when using mender which is then available in the UBoot's environment **mender_dtb_name** variable.

2. From what I've read in [the toradex forum](https://www.toradex.com/community/questions/5019/booting-to-wrong-dtb-file.html), it looks like there is no runtime board detection in place.

> You are correct, there are 4 device trees for 4 different possible configurations. For whichever you decide to use, you must set the fdt_file variable in the U-boot environment accordingly.

It is a rather old post, things may have changed, but even then, from the bbwarn in 1., mender considers only one dtb.

Am I overlooking something?

drewmoseley:
Yeah, I think that post is out of date. I just booted a colibri-imx6ull with the Toradex provided Reference Minimal Image which is dunfell based and the fdtfile is set using preboot dynamically:

```
Colibri iMX6ULL # printenv preboot
preboot=setenv fdtfile imx6ull-colibri${variant}-${fdt_board}.dtb
```

I think the best approach is to just ignore the warning from mender-helpers and then use  MENDER_UBOOT_PRE_SETUP_COMMANDS to make sure it's done dynamically.  The current beaglebone layer [does this here](https://github.com/mendersoftware/meta-mender-community/blob/dunfell/meta-mender-beaglebone/recipes-bsp/u-boot/u-boot_%25.bbappend) for reference.

mrpelotazo:
> ... use MENDER_UBOOT_PRE_SETUP_COMMANDS to make sure it's done dynamically. 

Using MENDER_UBOOT_PRE_SETUP_COMMANDS now. The change to the preboot variable has been reverted to its original value.

drewmoseley:
That still hard codes it.  What are you trying to accomplish here? What is the dynamic setting not doing that you need?

mrpelotazo:
>  That still hard codes it.

The way it is now **fdtfile** ends up being set to **imx6ull-colibri-eval-v3.dtb**. **KERNEL_DEVICETREE** has no effect in UBoots env besides of setting the **mender_dtb_name** variable which is not being used anywhere, so **fdtfile** remains with the above mentioned value. One has to make changes manually in the UBoot env to set something different.

> What is the dynamic setting not doing that you need?

What do you mean by dynamic setting? From what I see in the [bbappend](https://github.com/mendersoftware/meta-mender-community/blob/dunfell/meta-mender-beaglebone/recipes-bsp/u-boot/u-boot_%25.bbappend) you posted and from what I'trying to do, there is no difference to me. Essentially **setenv** is being used to set some variable to another variable's value.

Moreover from what you wrote:

>  I just booted a colibri-imx6ull with the Toradex provided Reference Minimal Image which is dunfell based and the fdtfile is set using preboot dynamically.
```
Colibri iMX6ULL # printenv preboot
preboot=setenv fdtfile imx6ull-colibri${variant}-${fdt_board}.dtb
```

What is dynamic about that? **imx6ull-colibri${variant}-${fdt_board}.dtb** evaluates to **imx6ull-colibri-eval-v3.dtb**. To get a different value one has to either set **variant** or **fdt_board** manually.

> What are you trying to accomplish here?

I'm just trying to get **KERNEL_DEVICETREE** (yocto) to be set to **fdtfile** (UBoot env).
I'm replacing the variables used to evaluate **fdtfile** value from **imx6ull-colibri${variant}-${fdt_board}.dtb** to **${mender_dtb_name}** which is set to **KERNEL_DEVICETREE**.

I hope this clarifies what I want to accomplish and how. If I'm misunderstanding something, please elaborate a little so I can make proper changes.

drewmoseley:
OK, sorry, that was _my_ misunderstanding. I somehow had the notion that variant, soc and friends were dynamically set by U-Boot itself but the Toradex BSP team has set me straight. Indeed the preboot method or the UBOOT_MENDER_PRE_SETUP_COMMAND method are identical.

If you want to just drop the latest 2 commits and get this set back to the original change, I think we can merge it.

mrpelotazo:
> OK, sorry, that was my misunderstanding.

No worries.

I've reverted commits to the original PR ones.

drewmoseley (approved):
LGTM

drewmoseley (approved):

drewmoseley:
@TheMeaningfulEngineer I guess I don't have write access to this repo so you'll have to merge this.

mrpelotazo:
@TheMeaningfulEngineer how is this looking?

TheMeaningfulEngineer:
Thanks, merging

PR #226, opened 2021-04-30
fix the state script

anishmonachan7:
The script is failing at different points and the script is fixed and tested.

### DIFF
--- a/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
+++ b/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
@@ -5,9 +5,9 @@ mnt=
 cleanup() {
     [ -n "$mnt" ] || return
     for d in sys proc dev run; do
-	if mountpoint -q "${mnt}/${d}"; then
-	    umount "${mnt}/${d}" >/dev/null 2>&1 || true
-	fi
+        if mountpoint -q "${mnt}/${d}"; then
+            umount "${mnt}/${d}" >/dev/null 2>&1 || true
+        fi
     done
     if mountpoint -q "$mnt"; then
         umount "$mnt" >/dev/null 2>&1 || true
@@ -17,7 +17,7 @@ cleanup() {
 
 echo "Installing NVIDIA bootloader update payload"
 
-num_slots=`nvbootctrl get-number-slots`
+num_slots=$(nvbootctrl get-number-slots)
 if [ $num_slots != 2 ]; then
     echo "Enabling A/B update mode using nv_update_engine"
     nv_update_engine --enable-ab
@@ -26,22 +26,27 @@ if [ $num_slots != 2 ]; then
         exit 1
     fi
 fi
-current_slot=`nvbootctrl get-current-slot`
-echo "Current boot slot: $current_slot"
 
-new_boot_part=`fw_printenv -n mender_boot_part`
-mnt=`mktemp -d -t nvbup.XXXXXX`
+current_boot_part=$(mount | grep ' / ' | cut -d' ' -f 1)
+required_boot_part=$(cat /etc/mender/mender.conf | sed 's/[,]  */&\n/g' | grep /dev/mmcblk0p | grep -v ${current_boot_part})
+new_boot_part=$(echo ${required_boot_part} | cut -d: -f2 | cut -d, -f1 | tr -d '" ')
+
+mnt=$(mktemp -d -t nvbup.XXXXXX)
+
 if [ -z "$mnt" -o ! -d "$mnt" ]; then
     echo "ERR: could not create directory for mounting install partition" >&2
     exit 1
 fi
-# Support delta updates by mounting the rootfs read only
-mount -o ro /dev/mmcblk0p${new_boot_part} "$mnt"
+
+# Support delta updates by mounting the rootfs read only if necessary
+mount ${new_boot_part} "$mnt"
+
 if [ ! -d "${mnt}/opt/ota_package" ]; then
     echo "ERR: Missing /opt/ota_package directory in installed rootfs" >&2
     cleanup
     exit 1
 fi
+
 # nv_update_engine needs access to these filesystems,
 # so bind-mount them into the new rootfs for the chroot
 mount --bind /sys "${mnt}/sys"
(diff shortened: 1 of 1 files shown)


meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
```diff
@@ -50,21 +55,31 @@ mount --bind /dev "${mnt}/dev"
...
-# Run the update engine in the context of the just-installed rootfs
-# to ensure that the TNSPEC in the config file it uses matches the
-# TNSPEC in the update payload. But first we have to set up the
-# configuration file with the TNSPEC.
-if [ -L "${mnt}/etc/nv_boot_control.conf" -a -x "${mnt}/usr/bin/setup-nv-boot-control" ]; then
-    mkdir -p "${mnt}/run/tegra-nv-bootctrl"
-    if ! chroot "${mnt}" /usr/bin/setup-nv-boot-control; then
-	echo "ERR: could not initialize nv_boot_control.conf in new rootfs" >&2
-    fi
+if [ -f "${mnt}/etc/nv_boot_control.conf" ]; then
```
dwalkes, line 58: @anishmonachan7 is the issue that you are updating from a rootfs which doesn't have the `nv_boot_control.conf` symlink setup [here](https://github.com/mendersoftware/meta-mender-community/blob/dunfell/meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-nv-boot-control-config_%25.bbappend#L2)?

I think we'll want to keep the logic which handles the symlink case so we could still support delta updates, however maybe we should have logic which can handle the case `${mnt}/etc/nv_boot_control.conf` exists, remount read write, and add a print which states delta updates aren't supported for this case.

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
```diff
@@ -50,21 +55,31 @@ mount --bind /dev "${mnt}/dev"
...
 fi
+
+# # Run the update engine in the context of the just-installed rootfs
+# # to ensure that the TNSPEC in the config file it uses matches the
+# # TNSPEC in the update payload. But first we have to set up the
+# # configuration file with the TNSPEC.
+
+cat ${mnt}${sysconfdir}/nv_boot_control.template >${mnt}${sysconfdir}/nv_boot_control.conf
+
+chmod 0644 "${mnt}${sysconfdir}/nv_boot_control.conf"
```
dwalkes, line 69: @anishmonachan7 do you understand why the top two lines aren't handled in [setup-nv-boot-control](https://github.com/OE4T/meta-tegra/blob/40701c99c42563a871b8fca0116440f0958a1124/recipes-bsp/tools/setup-nv-boot-control/setup-nv-boot-control.sh.in) for this specific case?

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
```diff
@@ -26,22 +26,27 @@ if [ $num_slots != 2 ]; then
         exit 1
     fi
 fi
-current_slot=`nvbootctrl get-current-slot`
-echo "Current boot slot: $current_slot"
 
-new_boot_part=`fw_printenv -n mender_boot_part`
-mnt=`mktemp -d -t nvbup.XXXXXX`
+current_boot_part=$(mount | grep ' / ' | cut -d' ' -f 1)
```
dwalkes, line 30: I think this can fail for some more complicated setups where rootfs is an overlay, not sure whether we need to support this.  Here's a way to do a simmilar thing with stat: https://github.com/OE4T/meta-tegra/issues/143#issuecomment-509917250

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
```diff
@@ -26,22 +26,27 @@ if [ $num_slots != 2 ]; then
         exit 1
     fi
 fi
-current_slot=`nvbootctrl get-current-slot`
-echo "Current boot slot: $current_slot"
 
-new_boot_part=`fw_printenv -n mender_boot_part`
```
dwalkes, line 32: I'm concerned setting `new_boot_part` which doesn't match mender settings may cause other issues.  I'm guessing you are in a situation where your boot/root slot alignment isn't proper due to issues like the ones discussed in https://github.com/BoulderAI/meta-mender-community/pull/1#issue-516955713 ?  It might be beter to try to detect/correct that problem instead of using something other than `mender_boot_part` to select the new boot partition.

dwalkes (changes requested):
Thanks @anishmonachan7 I've added some comments here.

Can you add more detail about what specifically is failing and how, and what upgrade sequence you are attempting?

Will I be able to reproduce the problems you've noticed on https://github.com/OE4T/tegra-demo-distro/tree/dunfell-l4t-r32.4.3 using `demo-image-base` and `PREFERRED_PROVIDER_virtual/bootloader = "u-boot-tegra"`?  I'll give that a try.

Also tagging @madisongh and @ichergui for their feedback.

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
```diff
@@ -26,22 +26,27 @@ if [ $num_slots != 2 ]; then
...
+
 if [ -z "$mnt" -o ! -d "$mnt" ]; then
     echo "ERR: could not create directory for mounting install partition" >&2
     exit 1
 fi
-# Support delta updates by mounting the rootfs read only
-mount -o ro /dev/mmcblk0p${new_boot_part} "$mnt"
+
+# Support delta updates by mounting the rootfs read only if necessary
+mount ${new_boot_part} "$mnt"
```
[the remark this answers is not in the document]
anishmonachan7: oh yes, This is some thing I've missed. I'll add it

anishmonachan7:
> Thanks @anishmonachan7 I've added some comments here.
> 
> Can you add more detail about what specifically is failing and how, and what upgrade sequence you are attempting?
> 
> Will I be able to reproduce the problems you've noticed on https://github.com/OE4T/tegra-demo-distro/tree/dunfell-l4t-r32.4.3 using `demo-image-base` and `PREFERRED_PROVIDER_virtual/bootloader = "u-boot-tegra"`? I'll give that a try.
> 
> Also tagging @madisongh and @ichergui for their feedback.

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot
anishmonachan7: yes, sure. I'll check that out

anishmonachan7:
@dwalkes Thank you for reviewing the pull request. I've encountered two issues with the scrip
1. The  `new_boot_part` and `current_boot_part` are outputting wrong value.
2. The `[setup-nv-boot-control](https://github.com/mendersoftware/meta-mender-community/blob/dunfell/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/redundant-boot-install-script-uboot#L57)` is failing with a message 'can't operate on dangling symlink'. 

I've discovered some new information from your comments. I'm closing the pull request because I'll test the alternative suggestion you have given for finding `new_boot_part` and `current_boot_part`. Also, to put back the support for delta update which is important.

PR #244, opened 2021-09-03
Toradex Apalis IMX8 display overlay support

pgiangrossi:
Update u-boot-distro-boot to support overlays, enabling display
support on the Apalis IMX8.

Signed-off-by: pagi <paul.giangrossi@qamcom.se>


drewmoseley:
Hi @pgiangrossi it's unclear to me where these files came from. I'm really wondering if we should do this as a patch file against the upstream Toradex BSP version of the file. I think that will make it clearer what we are changing and why.  As well, it will more easily allow us to detect when changes occur in the upstream BSP that might be needed as the patch will likely fail in that case and we can investigate.  I'll take a swing at that so you can see what I mean.

drewmoseley:
These changes are part of [this PR](https://github.com/mendersoftware/meta-mender-community/pull/246).  Closing

PR #247, opened 2021-09-29
Support Jetpack 4.6

jchillerup:
For this to work with OpenEmbedded4Tegra, as per the [release notes for L4T 32.6.1](https://github.com/OE4T/meta-tegra/wiki/L4T-R32.6.1-Notes#flash-layout-file-changes-for-t194-xavier-platforms)

> The flash layout XML files for the Xavier platforms in the L4T BSP have been modified to include new attributes (but no location/size changes). In particular, the xusb-fw partition is now marked oem_signed="true", as cboot now performs signature validation on the USB controller firmware.

I've added that in the file to have `meta-mender-community` still build for Jetpack 4.6 targets.

### DIFF
--- a/meta-mender-tegra/recipes-bsp/tegra-binaries/mender-custom-flash-layout/jetson-xavier-nx-devkit-emmc/flash_mender.xml
+++ b/meta-mender-tegra/recipes-bsp/tegra-binaries/mender-custom-flash-layout/jetson-xavier-nx-devkit-emmc/flash_mender.xml
@@ -214,7 +214,7 @@
             <description> **Required.** Slot B; contains slot status for A/B boot and A/B
               update. </description>
         </partition>
-        <partition name="xusb-fw" type="xusb_fw">
+        <partition name="xusb-fw" type="xusb_fw" oem_sign="true">
             <allocation_policy> sequential </allocation_policy>
             <filesystem_type> basic </filesystem_type>
             <size> 196608 </size>
@@ -226,7 +226,7 @@
             <description> **Required.** Slot A; contains XUSB module’s firmware file, making XUSB
               a true USB 3.0 compliant host controller. </description>
         </partition>
-        <partition name="xusb-fw_b" type="xusb_fw">
+        <partition name="xusb-fw_b" type="xusb_fw" oem_sign="true">
             <allocation_policy> sequential </allocation_policy>
             <filesystem_type> basic </filesystem_type>
             <size> 196608 </size>


TheMeaningfulEngineer:
Hey @dwalkes 
any ideas on this PR?

Hey @jchillerup 
can you please rebase on the latest dunfell.
Also @dwalkes is the community maintainer for `meta-mender-tegra` he'll make the call on how to approach this.

dwalkes:
@TheMeaningfulEngineer @jchillerup sorry I'm not sure how I didn't notice this originally.

My concern is that we've used the dunfell branch here for both the 32.4.3 branch at https://github.com/OE4T/tegra-demo-distro/tree/dunfell-l4t-r32.4.3 and https://github.com/OE4T/tegra-demo-distro/tree/dunfell so I think we need to be careful about any change for JP4.6/r32.6.1 specifically.

I'm also a bit confused because I believe we've successfully used the branch at https://github.com/OE4T/tegra-demo-distro/tree/dunfell on JP4.6 without this change, so I need to understand why we didn't see this issue.  Let me try to get caught up on this, and if you have any more details about testing with r32.4.3 please add here.

dwalkes:
After further investigation this morning, I think the issue is that we have dynamic layers in https://github.com/OE4T/meta-tegra-community/tree/dunfell/dynamic-layers/meta-mender-tegra which already include the changes here, for use with Jetpack 4.6 and the `dunfell` branch on `meta-tegra` which has since been upgraded from Jetpack 4.4 to Jetpack 4.6.

We have a separate set of layers in https://github.com/OE4T/meta-tegra-community/tree/dunfell-l4t-r32.4.3/dynamic-layers/meta-mender-tegra which don't include changes here and are intended to target Jetpack 4.4, the original meta-mender-community dunfell branch target.

I think there are two options:
1) We could move the dynamic layer content from https://github.com/OE4T/meta-tegra-community/tree/dunfell-l4t-r32.4.3/dynamic-layers/meta-mender-tegra here, close this PR, and recommend that anyone who needs Jetpack 4.6 support on dunfell add in the dynamic layers from https://github.com/OE4T/meta-tegra-community/tree/dunfell/dynamic-layers/meta-mender-tegra.
2) We could move the dynamic layer content from https://github.com/OE4T/meta-tegra-community/tree/dunfell/dynamic-layers/meta-mender-tegra here and retarget the `dunfell` branch to Jetpack 4.6 instead of Jetpack 4.4, updating https://hub.mender.io/t/nvidia-tegra-jetson-tx2/123 and other pages to reference the dunfell branch of https://github.com/OE4T/tegra-demo-distro instead of the `dunfell-l4t-r32.4.3` branch.

The problem with option 1 is that anyone who just wants the latest content from NVIDIA won't get it.  We could possibly remedy this with some updates to the mender hub pages like https://hub.mender.io/t/nvidia-tegra-jetson-tx2/123

The problem with option 2 is that anyone who is already using jetpack 4.4 and doesn't intend to update will have their branch automatically update when pulling from the latest meta-mender-community.

My thought is that most people using this layer with tegra will want to be aware of and probably start with tegra demo distro anyway, so the fact that we keep dynamic layers there which are required for newer branches/Jetpack releases is less of an issue and can be solved with better documentation, especially notes in https://hub.mender.io/t/nvidia-tegra-jetson-tx2/123.  This is making me think option 1 is a better option.

I appreciate any thoughts or feedback about the best way forward here.

dwalkes:
@jchillerup @TheMeaningfulEngineer I'd like to close this PR and merge https://github.com/mendersoftware/meta-mender-community/pull/254 instead.  Let me know if you have questions or concerns based on the discussion there.

PR #251, opened 2021-10-12
toradex: Support both 5.3.0 and 5.4.0 BSPs in Dunfell.

drewmoseley:
Toradex uses the Yocto Dunfell branch for all the 5.x releases of their BSP.
Currently the only changes needed between 5.3 and 5.4 are refreshed patches for u-boot-toradex
and u-boot-distro-boot.  Rather than complicate things at the branch level, requiring ongoing
maintenance of multiple branches, we will just set the default to 5.3.0 but allow users
to override by setting the TORADEX_BSP_VERSION bitbake variable.

Changelog: Title


kacf:
I don't fully understand why this is better than just normal splitting of bbappend files into two versioned recipes. Both patch variants are still in the layer, so what is saved by doing it the way you have proposed? Just thinking that using bbappends is more standard and less surprising.

drewmoseley:
How would I conditionally apply the BSP-version-specific patches with bbappends? Both toradex branches use u-boot-toradex_2020.07.bb but just a newer minor release tag in the newer toradex branch.

kacf (approved):
Ok, I see the problem. This is a good workaround then!

PR #255, opened 2022-02-05
Add support for verdin-imx8mm module on Toradex BSP 5.5.0

jsolla:


TheYoctoJester:
Hi @jsolla,

I'm really sorry it took me so long to get back to this PR. Thank you for the contribution! Just a few minor things that I need fixing.

1) can you please add your sign-off?
2) this does essentially not add support for a specific board, but modifies the exiting one to be compatible with the Toradex distribution, right? What does this mean for usage without the distribution, or with an older version? I obviously cannot test it here, but it would be good to document the intent and its side effects in the Readme.

Thanks!

jsolla:
Hi  @TheYoctoJester 
1.  Added signature
2. This commit doesn't modify existing support, it just adds support for manufacturer (Toradex) BSP version 5.5. Older versions remain untouched as decision to select older versions of the BSP is made by declaring variable ${TORADEX_BSP_VERSION} in local.conf so the recipe would take the files from separate BSP dirs.

In short: just followed the convention of the original recipe and added just a new directory with patches for newer BSP.

TheYoctoJester (approved):

PR #258, opened 2022-03-22
Change repo sync manifest form ssh to https

jarrocha:
Changelog: There was a recent change to the repo configuration for metal atmel
and does not allows ssh connections, only https.

Signed-off-by: Jaime Arrocha <jarrocha.dev@gmail.com>


TheYoctoJester (approved):
The change of the `poky` URI is not necessary, but doesn't hurt as far as I can see.

PR #259, opened 2022-03-31
Speed up installation

drewmoseley:
These two changes make a huge difference in some cases. I have an image that I am installing using Toradex Easy Installer on an apalis-im6. Before these changes, the install took about 30 min.  After this change it was about 6 min.

### DIFF
--- a/meta-mender-toradex-nxp/classes/image_type_mender_tezi.bbclass
+++ b/meta-mender-toradex-nxp/classes/image_type_mender_tezi.bbclass
@@ -2,11 +2,11 @@
 # Based largely on meta-toradex-bsp-common/classes/image_type_tezi.bbclass
 
 # Make sure the full Mender multi-partition image is available
-IMAGE_TYPEDEP_mender_tezi_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', ' sdimg.bz2', '', d)} \
-                                       ${@bb.utils.contains('IMAGE_FSTYPES', 'ubimg', ' ubimg.bz2', '', d)} \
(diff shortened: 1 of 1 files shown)


drewmoseley:
@TheYoctoJester @lluiscampos can you guys give this a quick review?

TheYoctoJester (approved):

ericrwx:
@drewmoseley: Great to see the speedups here!

With the removal of `bz2` compression, the resulting raw image may be quite large (e.g. an 8GB raw vs. 100MB compressed).  This may introduce new bottlenecks in deployment.  Perhaps `gz` strikes a better balance between _decompression speed_ and _compression ratio_?

drewmoseley:
Hi @ericrwx would it help if we compressed the entire mender_tezi.tar file instead of the sdimg payload? It will give you the speed up as well as the reduced size of the tarball.  The downside is if you extract the tarball to a TEZI server then it will be the large file but usually that's a local-area connection so may be a reasonable tradeoff.

drewmoseley:
@ericrwx [see here](https://github.com/mendersoftware/meta-mender-community/pull/260) for my proposal

ericrwx:
> would it help if we compressed the entire mender_tezi.tar file instead of the sdimg payload? It will give you the speed up as well as the reduced size of the tarball.

IIRC, the TEZI server uses the unpacked tezi.tar files, no?  So compressing mender_tezi.tar doesn't impact unattended flashing speeds.

> The downside is if you extract the tarball to a TEZI server then it will be the large file but usually that's a local-area connection so may be a reasonable tradeoff.

Indeed.  For larger flash sizes (e.g. 8GB or more), the sdimg file transfer time may start to dwarf the bz2 decompression time, especially on slower transports like 100 Mbit LAN or USB 2.0.  Therefore, it seems preferable to keep some type of sdimg compression.

PR #262, opened 2022-04-03
tegra: add symlinks for L4T R32.7.1 flash layout, u-boot patches

madisongh:
The flash layouts and u-boot patches for R32.6.1 also work with L4T R32.7.1, so this commit just adds symlinks named for the updated BSP to point to the existing directories.


dwalkes:
Thanks @madisongh !  @TheYoctoJester this is good to merge from my perspective.

MuchoLucho (approved):
It looks good to me

PR #269, opened 2022-05-18
Toradex Kirkstone

drewmoseley:
Note that this barely works.  Since Toradex does not yet have BSP support for Kirkstone we can only test with the meta-freescale-3rdparty machines which does not include all of the Toradex machines.  The only machines I have tested here are:

* apalis-imx6: build and boot tested
* colibri-imx6ull: build tested

Also, until Toradex has a Kirkstone BSP, the Toradex Easy Installer is unusable on this branch so getting these things installed requires a lot of `dd` work and I don't really plan to document it.  This should at least get the branch building and ready and when Toradex publishes official support we can revisit.


TheYoctoJester:
@drewmoseley thanks a lot for the submission! I can merge it as-is if you are okay with it. Given your comments, what do you think of this:
- not removing the sanity check class and inheritance
- replacing the bbfatal with a bbwarn that outlines the state of the support
That you users would get a notification that explains that this should be seen as preliminary, but they can still use it?

drewmoseley:
@TheYoctoJester sounds reasonable. I updated the warning and force-pushed the updated branch. This is RTG.

PR #284, opened 2022-10-31
toradex: Use mender_dtb_name consistently.

drewmoseley:
Make sure we use the mender_dtb_name variable consistently across all support Toradex platforms. This loses the ability to define it dynamically which is the default (sort of) for Toradex but is better inline with standard Mender/Yocto integrations.

Changelog: Title

### DIFF
--- a/meta-mender-toradex-nxp/recipes-bsp/u-boot/files/toradex-bsp-5.6.0/apalis-imx6/0001-configs-toradex-board-specific-mender-integration.patch
+++ b/meta-mender-toradex-nxp/recipes-bsp/u-boot/files/toradex-bsp-5.6.0/apalis-imx6/0001-configs-toradex-board-specific-mender-integration.patch
@@ -1,12 +1,12 @@
-From d89ac1a7f228f957e22ccc2d0421d379f1650f49 Mon Sep 17 00:00:00 2001
+From 3e9f1c1bb6d32e5087aead5473e4e21eaeaccddc Mon Sep 17 00:00:00 2001
 From: pagi <paul.giangrossi@qamcom.se>
 Date: Fri, 9 Jul 2021 08:29:09 +0200
 Subject: [PATCH] 0001-configs-toradex-board-specific-mender-integration
 
 ---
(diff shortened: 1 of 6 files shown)


drewmoseley:
This commit makes the Toradex integrations much more consistent with Yocto/Mender standards but it does potentially result in behavior changes as this forces the builds to boot with the DTB specified by the mender_dtb_name uboot environment variable. The colibri-imx6ull integration already had this but the others did not.

I am interested in feedback from any users of the Toradex integrations on this.

cc: @TheMeaningfulEngineer @TheYoctoJester @jsolla @leon-anavi @lluiscampos @mirzak @rickyricksanchez @pgiangrossi

drewmoseley:
Also, I realize this is for dunfell. I am still working on kirkstone stuff so I'll probably eventually get a version of this posted for that branch as well.

TheYoctoJester (approved):
Basically I am fine with the PR. Can you please confirm that the uboot config changes that are also contained are intentional, then I'll merge. @drewmoseley

drewmoseley:
The u-boot changes are the meat of this PR. Basically it sets the fdtname U-Boot environment variable to the value specified by the MENDER_DTB_NAME.

But since this changes functionality, maybe we should wait a few more days for others to chime in.

TheYoctoJester:
Yeah I understand that its all about u-boot, but there are a bunch of configuration changes like `CONFIG_FIT_VERBOSE=y` that change behaviour but don't seem to be related to the dtb handling, hence the question if those accidentially slipped in.

drewmoseley:
I don't _think_ those are actual changes; I think it's an artifact of the "diff of a diff" view since we are actually committing patches. The context around the modified lines is actually unmodified but since the new version has a new modified line, additional context is pulled into the patch.  At least I _think_.  Patch inception makes my head hurt.

TheYoctoJester:
I'll merge then to hopefully solve the headaches.

TheYoctoJester:
I'll merge then to hopefully solve the headaches.

TheYoctoJester:
I'll merge then to hopefully solve the headaches.

PR #285, opened 2022-11-07
meta-mender-nxp: add kirkstone support for Boundary Devices

chrisdimich:


TheYoctoJester:
Hi @chrisdimich,

Thanks for the integration refresh! Just having given it a quick skim yet, I noticed that the machine specific overrides in the `local.conf.template` still use the old syntax, and will supposedly not kick in. Is this intentional?

Greetz

chrisdimich:
@TheYoctoJester I was never managing those machines, so didn't worry about them. I only did this support for Boundary Devices. Is that okay?

TheYoctoJester:
Hi @chrisdimich yes I understand. I would like to give this a build check to make sure there are no syntax clashes, and then I'll merge. Is that okay with you? Will take a couple of days, I'm travelling at the moment, or do you have an immediate need to get it included?

chrisdimich:
No worries thanks! Whenever you get to it is fine. Appreciate the quick response!

TheYoctoJester:
@chrisdimich Needed a little sorting out in our repo but looks fine now, thanks. One minor observation:
- the u-boot patches apply only with fuzz, a refresh would be good

TheYoctoJester:
@chrisdimich sorry to be nagging, but the Readme mentions
- nitrogen8m
- nitrogen8mm
- nitrogen8mm (duplicate!)
- nitrogen8mq
whereas in the head of the `local.conf.append` template and the corresponding/linked Mender Hub article only the nitrogen8m is mentioned.

Can you prepare a follow-up here to get those into a consistent state, maybe?

Greetz,
Josef

TheYoctoJester:
The kirkstone branch was indeed created after the PR. I will definitely carry the changes over.

chrisdimich:
@TheYoctoJester Thanks! Also sorry missed your note -> "the u-boot patches apply only with fuzz, a refresh would be good."

I will look into that.

PR #288, opened 2022-11-11
Toradex bsp6 kirkstone

drewmoseley:
This set of patches cleans up all the old BSP 5 bits, and gets everything working with Toradex BSP 6, at least insofar as modules are supported (apalis-imx8 and colibri-imx8 are explicitly not supported by Toradex yet).


drewmoseley:
Wrong base branch.

PR #293, opened 2022-11-13
meta-mender-nxp: imx7s-warp: rebase patch on kirkstone

texierp:
Also update MENDER_STORAGE_DEVICE entry.

Tested with the following sequence:

root@imx7s-warp:~# mender install  /data/core-image-base-imx7s-warp.mender INFO[0000] Loaded configuration file: /var/lib/mender/mender.conf INFO[0000] Loaded configuration file: /etc/mender/mender.conf INFO[0000] Mender running on partition: /dev/mmcblk2p2 INFO[0000] Start updating from local image file: [/data/core-image-base-imx7s-warp.mender] Installing Artifact of size 50458112...
INFO[0000] No public key was provided for authenticating the artifact WARN[0000] Returning artifact name from /etc/mender/artifact_info file. This is a fallback, in case the information can not be retrieved from the database, and is only expected when an update has never been installed before. INFO[0000] Opening device "/dev/mmcblk2p3" for writing INFO[0000] Native sector size of block device /dev/mmcblk2p3 is 512 bytes. Mender will write in chunks of 1048576 bytes .............................................................  -  99 %INFO[0274] All bytes were successfully written to the new partition INFO[0274] The optimized block-device writer wrote a total of 929 frames, where 166 frames did need to be rewritten (i.e., skipped) INFO[0276] Wrote 973078528/973078528 bytes to the inactive partition .............................................................. - 100 % INFO[0276] Enabling partition with new image installed to be a boot candidate: 3 Use 'commit' to update, or 'rollback' to roll back the update. At least one payload requested a reboot of the device it updated.

Signed-off-by: Pierre-Jean Texier <texier.pj2@gmail.com>


TheYoctoJester (approved):
Thanks @texierp! Just a short question, why is the MENDER_STORAGE_DEVICE entry change needed? Do you happen to know the root cause?

texierp:
Thanks @TheYoctoJester 

> Thanks @texierp! Just a short question, why is the MENDER_STORAGE_DEVICE entry change needed? Do you happen to know the root cause?

Yes, depending on the Kernel version, the mmc block number may change as the mmc block number is not determistic :/

PR #299, opened 2022-11-16
RFC toradex: Move dtb and dtbo files to root partition.

drewmoseley:
These files should be loaded from the rootfs partition so we can update them OTA.

Changelog: Title

### DIFF
--- a/meta-mender-toradex-nxp/classes/mender-toradex.bbclass
+++ b/meta-mender-toradex-nxp/classes/mender-toradex.bbclass
@@ -9,6 +9,16 @@ toradex_mender_update_fstab_file() {
     mv ${IMAGE_ROOTFS}${sysconfdir}/fstab.toradex ${IMAGE_ROOTFS}${sysconfdir}/fstab
 }
 
+ROOTFS_POSTPROCESS_COMMAND:append = " toradex_mender_update_devicetree_overlays;"
+toradex_mender_update_devicetree_overlays() {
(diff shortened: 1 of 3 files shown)


drewmoseley:
@TheYoctoJester @leon-anavi I would love your input on this. I noticed that we were not pulling DTB and DTBO files from the rootfs partition so there was no way to update them OTA.  This patch adds that capability. This was discovered while trying to do an OTA from dunfell to kirkstone and is required for that ability. I have an equivalent patch on kirkstone that I am working on but getting it into dunfell is a necessary first step.

TheYoctoJester:
@drewmoseley yeah we've run into a similar situation. I personally think that this looks good and is the correct approach. From a version management perspective, `kirkstone` is meant to take the lead now on this repo.

drewmoseley:
About using kirkstone, understood but this is required if we want to allow OTA between dunfell and kirkstone so I figured we start here. I have the equivalent fix for kirkstone in my WIP branch locally and will submit that once we make a final decision on this one.

drewmoseley:
Let's wait for @leon-anavi; he's the only other active user that I am aware of.

leon-anavi:
Sounds as a very good idea and the patch looks good to me. 

Thanks, Leon

drewmoseley:
OK.  In that case I guess we are good to go.

TheYoctoJester (approved):

PR #306, opened 2022-11-22
meta-mender-raspberrypi: extend boot partition to 64MByte

TheYoctoJester:
Since kirkstone, the olde default boot patition size of 40MByte is not sufficient anymore for the Raspberry Pi 3.

Ticket: None
Changelog: None

Signed-off-by: Josef Holzmayr <jester@theyoctojester.info>

### DIFF
--- a/meta-mender-raspberrypi/templates/local.conf.append
+++ b/meta-mender-raspberrypi/templates/local.conf.append
@@ -4,7 +4,7 @@
 MACHINE ?= "raspberrypi3"
 
 RPI_USE_U_BOOT = "1"
-MENDER_BOOT_PART_SIZE_MB = "40"
+MENDER_BOOT_PART_SIZE_MB = "64"
 IMAGE_INSTALL:append = " kernel-image kernel-devicetree"
 IMAGE_FSTYPES:remove = " rpi-sdimg"
 


tranchitella (approved):

daqbroker:
I'm not even beginning to suggest that this is wrong, especially as it relates to something that already exists and please forgive me if I am committing some sort of yocto hate crime by posting here.

I have a fleet of raspberry pis in the field (mix of rpi3 and rpi4) which install mender updates build with a `local.conf` file that currently is mostly the default of the `setup-environment` script. Hence on the `dunfell` branch, my current value for `MENDER_BOOT_PART_SIZE_MB` is `40`.

If I understand correctly, this change of default value makes `dunfell` devices impossible to migrate to `kirkstone` as they stand due to the partition being 'fixed' at 40 MB. This is what I am seeing both when running `bitbake` at 40MB of boot partition size (the compilation fails with python scripts complaining out of disk space) and when running mender update on an existing device with a boot partition of 64MB (a `cp` problem with 'No space left on device' is also triggered).

Is there some sort of migration magic that I am not aware of or something that could be done from my side? The only solution I see is running a script on all these devices to "manually" reclaim space from the `/data` partition (which is auto inflating to the whole size of the SD card).

More than happy to provide any mender/yocto configuration and logs required (I'm assuming no one wants to dive into what I have so I'm not providing it here, I'm sure I've more than understood the problem here).

Thanks in advance

António Dias (amcbd89@gmail.com)

TheYoctoJester:
Hi @daqbroker,

No worries, no hate detected here. You're right that this *might* be problematic if you start out with a 40MB `/boot` from `dunfell`. And yes again, the solution that you describe is kinda the nuclear one: repartitioning.

However, its by far not as bad as it seems. The partitioning mostly affects the initial image. The contents of the boot partition is not touched/extended during a standard update, so at least ad hoc I do not see any particular reason why this should fall over. The one thing to be aware of is the needed `rpi-firmware-update` on the migration update as described here: https://docs.mender.io/troubleshoot/yocto-project-runtime#deploying-an-update-from-dunfell-to-kirkstone-on-raspberry-pi-bl

Beyond that, two measures that I advise:
- extend `/boot` to 64MB or even more for new devices.
- maybe have a look at the storage usage on `/boot` chances are that you don't need all of the stuff there, especially not the overlays, so you can trim it down for your usecase on `kirkstone`.

If possible, then it would be good to continue the discussion on the Mender Hub (hub.mender.io), because thats where most people look for solutions in case of problems and therefore the most benefit. Thanks!

Greetz,
Josef

daqbroker:
Thanks @TheYoctoJester!

I've created [a topic](https://hub.mender.io/t/dunfell-to-kirkstone-migration-boot-partition-on-flahsed-devices/5528) on mender hub to continue the discussion there. Linking here for anyone else interested in the discussion

PR #321, opened 2023-02-27
meta-mender-octavo-osd32mp: remove manifest default

TheYoctoJester:


oleorhagen (approved):
Why remove the `default remote` ? Just readability or is it deprecated somehow?

PR #322, opened 2023-02-28
enable ethernet interface on osd32mp1-red

amsobr:
Unfortunately I missed an important feature in the osd32mp1 integration: the ethernet port is not enabled (the original work used only wifi).

Please consider this PR to fix that issue and enable board connectivity using ethernet.

With this fix the board becomes able to connect to the internet and run OTA updates from hosted.mender.io

Thanks,
António


TheYoctoJester (approved):

TheYoctoJester:
> Unfortunately I missed an important feature in the osd32mp1 integration: the ethernet port is not enabled (the original work used only wifi).

No problem, such happens and nice you found it so quickly!

PR #323, opened 2023-03-01
Renesas RZG2L board integration

estape11:


scripts/mender.xml
```diff
@@ -3,9 +3,10 @@
   <default sync-j="4" revision="dunfell"/>
 
   <remote fetch="https://github.com/mendersoftware" name="mender"/>
+  <remote fetch="https://github.com/estape11" name="mender-temp"/>
```
TheYoctoJester, line 6: This needs to be removed before merging.

TheYoctoJester (changes requested):

scripts/mender.xml
estape11: Removed :D

TheYoctoJester (approved):

PR #331, opened 2023-03-14
meta mender fixes

fboudra:


meta-mender-intel/scripts/manifest-intel.xml
```diff
@@ -1,12 +1,13 @@
-<?xml version="1.0" encoding="UTF-8" ?>
-  <manifest>
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+  <remote fetch="git://git.yoctoproject.org" name="yocto"/>
+  <remote fetch="git://git.openembedded.org" name="oe"/>
 
-  <remote fetch="git://git.yoctoproject.org"        name="yocto"/>
+  <default revision="kirkstone" sync-j="4"/>
```
TheYoctoJester, line 6: I think this will clash with the upper level `scripts/mender.xml`, repo always gives me an error if there are two `<default/> statements.

meta-mender-intel/templates/local.conf.append
```diff
@@ -3,11 +3,10 @@
...
 MACHINE ?= "intel-corei7-64"
 
-# Note that these settings have been verified for booting from a SD card.
-# You might need to change the two latter settings if you are planning to
-# boot from a different storage medium.
+# Note that this setting has been verified for booting from a SD card.
+# You might need to change the setting if you are planning to boot from
+# a different storage medium.
 MENDER_STORAGE_DEVICE = "/dev/mmcblk2"
-MENDER_GRUB_STORAGE_DEVICE = "hd0"
```
TheYoctoJester, line 10: Commenting out would be nicer, so the variable is still visible.

TheYoctoJester (changes requested):

meta-mender-intel/templates/local.conf.append
fboudra: The variable is deprecated, it's causing a fatal error if used ("this is now dynamically determined at runtime."), isn't exposed in the documentation anymore, and it isn't in any other layer.

It seems to me a bad idea as you can't use it in any case. Why would you keep it commented?

meta-mender-intel/scripts/manifest-intel.xml
fboudra: How do you use the top level `scripts/mender.xml` ?

Following the instructions and documentation, it doesn't trigger any issues:
```
    $ mkdir mender-<vendor/soc name>
    $ cd mender-<vendor/soc name>
    $ repo init \
           -u https://github.com/mendersoftware/meta-mender-community \
           -m meta-mender-<vendor/soc name>/scripts/manifest-<vendor/soc name>.xml \
           -b kirkstone
    $ repo sync
```
and in meta-mender-intel:
```
mkdir mender-intel && cd mender-intel
repo init -u https://github.com/mendersoftware/meta-mender-community \
           -m meta-mender-intel/scripts/manifest-intel.xml \
           -b kirkstone
repo sync
source setup-environment intel
bitbake core-image-base
```

In addition, if I grep in the repository, there's 6 other layers using the default.

IMO, either it's removed on all the others as well, or we can it repo's wide. Opinion?

meta-mender-intel/templates/local.conf.append
TheYoctoJester: Ah you're right, sorry misremembered it. Removing it is fine then, of course.

meta-mender-intel/scripts/manifest-intel.xml
fboudra: @TheYoctoJester what about the above?

TheYoctoJester (approved):

PR #339, opened 2023-04-19
update-modules: remove kirkstone warnings

TheYoctoJester:


meta-mender-update-modules/conf/layer.conf
```diff
@@ -11,4 +11,4 @@ BBFILE_COLLECTIONS += "mender-update-modules"
 BBFILE_PATTERN_mender-update-modules = "^${LAYERDIR}/"
 BBFILE_PRIORITY_mender-update-modules = "15"
 
-LAYERSERIES_COMPAT_mender-update-modules = "kirkstone"
+LAYERSERIES_COMPAT_mender-update-modules = "kirkstone"
```
kacf, line 14: I guess it's fine, but what was the warning? That the file ends in a newline? Isn't that what it *should* end with?

kacf (approved):

meta-mender-update-modules/conf/layer.conf
oleorhagen: It's the opposice, no? THat it does _not_ end in a newline?

TheYoctoJester: Good catch indeed, yet, where did you get a warning?

kacf: I'm very confused now. Removing the newline was the only change in this PR...

PR #341, opened 2023-04-27
meta-mender-ti: add initial support

TheYoctoJester:
This adds initial support for boards maintained in meta-ti. The first board integration is the BeaglePlay platform.


kacf (approved):
Nice 👍

PR #342, opened 2023-04-28
Rockchip kirkstone updates

nandra:


nandra:
@TheYoctoJester please pull in latest fixes for rockpi-e board which was found during testing. I'll create board integration item soon. Thanks.

PR #343, opened 2023-05-01
toradex: Add support for BSP 5.7.2.

drewmoseley:
This adds support for the latest point release of the Toradex BSP.


drewmoseley:
@leon-anavi you may want to give this a shot.  A basic build/boot test worked for me but I don't have anything much more complicated than that.

Also, there is something wrong with the verdin-imx8mm support.  For some reason, the UBoot environment is missing all the Mender variables.  I don't see anything obviously wrong with the code and the patches seem to be applying properly but they just don't show up in the binary. I tried the 5.6.0 BSP and it seems to have the same issue so I have no idea how long it has been like that.  I'm not sure I ever tested that platform.

TheYoctoJester (approved):

leon-anavi:
> @leon-anavi you may want to give this a shot. A basic build/boot test worked for me but I don't have anything much more complicated than that.

Sorry for the late reply, I was pretty busy in the past days. I will give it a try next week on Dahlia with verdin-imx8mp.

> Also, there is something wrong with the verdin-imx8mm support. For some reason, the UBoot environment is missing all the Mender variables. I don't see anything obviously wrong with the code and the patches seem to be applying properly but they just don't show up in the binary. I tried the 5.6.0 BSP and it seems to have the same issue so I have no idea how long it has been like that. I'm not sure I ever tested that platform.

I don't have verdin-imx8mm to test this.

Best regards,
Leon

PR #350, opened 2023-07-04
meta-mender-tegra updates for L4T R32.7.4

madisongh:
Reworks the L4T version mapping to eliminate having to do an update here for every new L4T release:
 * limits L4T version specifics to the two recipes that need them
 * U-Boot patches have been compatible with every L4T release since R32.6.1, and that's not likely ever to change again
 * Likewise with the custom flash layouts

Still keeps compatibility with older L4T releases that may still be getting used.


dwalkes (approved):
Thanks @madisongh !

TheYoctoJester (approved):

PR #351, opened 2023-07-05
meta-mender-toradex-nxp: fix marketing and data location

fboudra:
Those files are deployed under ${DEPLOYDIR}/mender-tezi-metadata/ directory. Update their path to generate an image-*.json working out of the box.

### DIFF
--- a/meta-mender-toradex-nxp/classes/image_type_mender_tezi.bbclass
+++ b/meta-mender-toradex-nxp/classes/image_type_mender_tezi.bbclass
@@ -129,8 +129,8 @@ python rootfs_mender_tezi_json() {
     data["release_date"] = datetime.strptime(d.getVar('SRCDATE'), '%Y%m%d').date().isoformat()
     data["prepare_script"] = "prepare.sh"
     data["wrapup_script"] = "wrapup.sh"
-    data["marketing"] = "marketing_mender_toradex.tar"
-    data["icon"] = "mender_toradex_linux.png"
+    data["marketing"] = "mender-tezi-metadata/marketing_mender_toradex.tar"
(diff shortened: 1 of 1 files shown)


fboudra:
@drewmoseley ^^^

drewmoseley:
LGTM.  @TheYoctoJester can you review and merge?

TheYoctoJester (approved):

zach-welch-aquabyte:
This change breaks the `mender_tezi` tarball.  I assume that it was made to facilitate installing from the deploy directory directly, but the command that creates the tarball (later in the `bbclass`) strips the paths from all of the files.  When extracted, the `image.json` contains the subdir path, so the installer produces a 404 error when trying to access those files.

drewmoseley:
@zach-welch-aquabyte I don't follow.  Can you post the file contents showing the issue?  Any chance you have a fix you can submit?

zach-welch-aquabyte:
This change adjusts the path names in the image.json file to use the
subdirectory where those files can be found in the deploy directory.
However the installation tarball that is created uses a flattened directory
structure, so the json file contents already were correct for the extracted
files. We have automation that extracts the tarball and serves that for the
installer, and it fails with a 404 when trying to access the files
referenced in the json.

The fix is either to revert this patch or change the tarball creation to
preserve the subdirectory for these files. The former is what I tested here
before making my previous comment, but the latter is probably nicer in that
in makes things more flexible for developers. I was not sure about the
cleanest way to diff the tarball, and I’m now out of office for a couple of
days.

drewmoseley:
Ahh I think I see. @fboudra can you comment? I think having the mender_tezi tarball work out of the box is the right thing here.  Maybe your approach will require some post processing and we revert this?

PR #354, opened 2023-07-12
feat: kas configurations for known boards

TheYoctoJester:
Add kas configurations for boards that are known to build respectively under maintenance to have a
reproducible way of building them.

The kas files do not follow branch names, but provide verified commit hashes to ensure a known-good state is kept available.

### DIFF
--- a/kas/colibri-imx6ull.yml
+++ b/kas/colibri-imx6ull.yml
@@ -1,7 +1,7 @@
 header:
   version: 11
   includes:
-  - kas/include/mender-base.yml
+  - kas/include/mender-full-ubi.yml
   - kas/include/nxp.yml
   - kas/include/toradex.yml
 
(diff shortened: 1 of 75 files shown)


kas/colibri-imx6ull.yml
```diff
@@ -0,0 +1,22 @@
...
+  - kas/include/mender-base.yml
+  - kas/include/nxp.yml
+  - kas/include/toradex.yml
+
+machine: colibri-imx6ull
+
+local_conf_header:
+  colibri-imx8x: |
+    INHERIT:remove = " mender-full "
+    INHERIT:append = " mender-full-ubi "
```
kacf, line 13: Although I realize it will add a lot of lines, is it perhaps better to include `INHERIT:append = " mender-full"` in all regular boards instead of assuming it's there already, and removing it here? It might be more self contained, in a sense. `remove` is often a sign that you have added too much to the default setup.

kacf (approved):
Amazing work with all these definitions!

I don't understand everything that's going on here, so treat my comment below as a suggestion only.

TheYoctoJester:
@kacf refactored the `mender-base`, also fixed some board builds along the way.

kacf (approved):
Great! 👍

PR #355, opened 2023-07-23
meta-mender-tegra: kas related fixes and updates 

TheYoctoJester:
This merges
- the Readme update to not use the repo based setup mechanism anymore, and moves to kas
- the fix of an unintentional layer dependency duplication, resulting in a non-reproducible floating revision.


TheYoctoJester:
@dwalkes to prepare `kirkstone` for some of the things you commented on. Thoughts?

dwalkes:
@TheYoctoJester looks good, I haven't given the kas config a test drive, will try to kick that off tonight, but probably no need to wait for that.

kas/include/tegra.yml
```diff
@@ -10,8 +10,6 @@ repos:
     refspec: kirkstone-l4t-r32.7.x
 
   meta-openembedded:
-    url: https://git.openembedded.org/meta-openembedded
-    refspec: kirkstone
```
kacf, line 14: Nitpick: The commit only says that this is removed, but not why (and it's not clear to me).

kacf (approved):

kas/include/tegra.yml
TheYoctoJester: Very good catch. Addressed in update, okay with it?

kacf: 👍

PR #363, opened 2023-08-01
fix: kas: machine type for raspberrypi4-64

TheYoctoJester:
The machine type was a copy error from the
raspberrypi4.yml file.

Changelog: Title
Ticket: None


TheMeaningfulEngineer (approved):
Haven't tested this but seems just like a typo fix.
Approving

PR #367, opened 2023-08-17
build on master/mickledore, add application updates demo

TheYoctoJester:
This combines the work on building the master/mickledore branch of the upstream repositories with the demo for `docker-compose` based application updates on a Raspberry Pi 4.


meta-mender-update-modules/conf/layer.conf
```diff
@@ -11,4 +11,4 @@ BBFILE_COLLECTIONS += "mender-update-modules"
 BBFILE_PATTERN_mender-update-modules = "^${LAYERDIR}/"
 BBFILE_PRIORITY_mender-update-modules = "15"
 
-LAYERSERIES_COMPAT_mender-update-modules = "kirkstone"
\ No newline at end of file
+LAYERSERIES_COMPAT_mender-update-modules = "mickledore"
```
kacf, line 14: Why all three branches on the others, and not here?

.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        board: [
+        target: [
```
kacf, line 27: This commit does not affect users, should it have `Changelog: None`?

.github/workflows/build.yml
```diff
@@ -58,4 +59,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: enter build dir and build
-        run: mkdir -p $BUILDDIR/${{ matrix.target }} && cd $BUILDDIR/${{ matrix.target }} && kas build ../../kas/${{ matrix.target }}.yml
\ No newline at end of file
+        run: mkdir -p $BUILDDIR/${{ matrix.target }} && cd $BUILDDIR/${{ matrix.target }} && kas build $GITHUB_WORKSPACE/kas/${{ matrix.target }}.yml
```
kacf, line 62: Same here, maybe `Changelog: None` for this commit.

kacf (approved):
I don't fully understand the docker-compose based UM example, but since it's only an example, I don't think I will invest time in it, and just trust you on it!

meta-mender-update-modules/conf/layer.conf
TheYoctoJester: That was a borderline decision, because the new Application Update Module requires `mickledore`. I'm fine with either for now, but as I expect the branch to move forwards over the next months and converge to the next LTS release, it is semi-temporary anyways. If you think marking it for all three, and accepting the UM to break for non-`mickledore` builds I can do that @kacf

kacf: No, that's a good call, I was just curious!

PR #370, opened 2023-10-24
client-only: add layer

TheYoctoJester:
The meta-mender-client-only layer enables the Mender client to be installed and operated without the full integration in place.

Changelog: Title
Ticket: None

### DIFF
--- a/meta-mender-client-only/recipes-mender/mender-client/mender-client_%.bbappend
+++ b/meta-mender-client-only/recipes-mender/mender-client/mender-client_%.bbappend
@@ -1,0 +1,16 @@
+do_configure:prepend() {
+	if [ -n "${MENDER_FEATURES}" ]; then
+		bbwarn "having both the Mender integration and the meta-mender-client-only layer enabled will cause unexpected effects!"
+	fi
+}
+
+do_compile:append() {
+    echo "device_type=${MACHINE}" > ${B}/device_type
+}
+
+do_install:append() {
+    install -m 755 -d ${D}/${localstatedir}/lib/mender
+    install -m 444 ${B}/device_type ${D}/${localstatedir}/lib/mender/
+}
+
+RDEPENDS:${PN} = " ca-certificates"
--- a/meta-mender-client-only/LICENSE
+++ b/meta-mender-client-only/LICENSE
@@ -1,0 +1,202 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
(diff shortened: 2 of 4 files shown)


kacf (commented):
Is this an attempt at using the client without `mender-image`? I guess then we will see how well it works in practice. There are probably a few kinks.

TheYoctoJester:
@kacf yes that exactly is the purpose, for some reasons:
- makes it easy to get started with Mender on Yocto without having to do the full integration
- allows for UMs to use other, existing mechanisms such as SWupdate, RAUC, ...

kacf:
Great, makes a lot of sense!

meta-mender-client-only/recipes-mender/mender-client/mender-client_%.bbappend
```diff
@@ -0,0 +1,16 @@
+do_configure:prepend() {
+	if [ -n "${MENDER_FEATURES}" ]; then
+		bbwarn "having both the Mender integration and the meta-mender-client-only layer enabled will cause unexpected effects!"
+	fi
+}
+
+do_compile:append() {
+    echo "device_type=${MACHINE}" > ${B}/device_type
```
kacf, line 8: This should use `MENDER_DEVICE_TYPE`.

kacf (commented):
Hmm, come to think of it: Should this even be its own layer? I mean, shouldn't we make these fixes in the core layer instead, so that they will just work?

meta-mender-client-only/recipes-mender/mender-client/mender-client_%.bbappend
TheYoctoJester: ´MENDER_DEVICE_TYPE` is not set without the `INHERIT += "mender-xxx"` addition, see other comment.

TheYoctoJester:
@kacf the key idea is to be as little intrusive as possible. I also thought about pouring this into a `MENDER_FEATURE` like `mender-standalone` or whatever, but in any case that requires more effort, respectively creates more friction and clashes in more places. This way, at least to get started, the process matches developer expectations: add a layer, add something to `IMAGE_INSTALL`. Thats why I'd like to go this route, at least for the first iteration.

meta-mender-client-only/recipes-mender/mender-client/mender-client_%.bbappend
kacf: Hmm, ok. Maybe we can consider changing this though. I don't see any harm in setting a variable, even if no features are enabled. And it would enable this to work in the expected way.

kacf:
I see the incentive, but I actually think this really belongs in the core layer. And the fixes that will inevitably result from this work are all very welcome, it's an overlooked area up until now.

I think if we just stay very disciplined, this shouldn't create a lot of friction. And by this I mean:

* For the non-integration case, when just installing the client, we don't require any `MENDER_FEATURES` to be turned on.
* When all features are off, I think it's better to not think of the image as "not a Mender image, there are definitely no rootfs updates", but rather "this is a custom image, we don't know if it has rootfs updates or not, so let's not assume anything".
* Due to the last point, we must be careful that all image changes outside the recipe itself and its files, are strictly kept out of the build when no features are turned on.
    * That means no symlinks to data partitions.
    * It also means no `device_type` file, though we could offer this file in a separate recipe, or as a `PACKAGECONFIG`. The reason this is important is that this file by its very nature lives on a persistent partition. But in the non-integration case, we don't know where this partition is, so we should not offer it out of the box unless you ask for it.

What do you think? Am I overthinking it, or is there something there?

TheYoctoJester:
@kacf You are not overthinking it, this is actually the reason why it took so long to come with this solution, and why I asked so many questions. The key conclusion I took is "in a perfect world, solution for this would go into `meta-mender-core`, but I'm not sure when the required effort could be spent". Because it also requires figuring out the places where the recipes are not clean right now.
Then, how to fine tune the client installation where applicable. Still use `MENDER_FEATURES`? The same ones? Or add a `mender-client-only` feature which means this and is mutually exclusive with `mender-image`?
Also some experimentation I did was adding a new `mender-client-only` package which then pulls in `mender-client`, as opposed to a `.bbappend`.

In the long run we should strive for it, I completely agree. But until that is done, better to have something not perfect but good enough, instead of nothing. :-)

kacf (approved):
> The key conclusion I took is "in a perfect world, solution for this would go into `meta-mender-core`, but I'm not sure when the required effort could be spent". Because it also requires figuring out the places where the recipes are not clean right now.

Sure, but is this any less work in a separate layer? 🙂

In any case, I certainly don't want to stand in the way of this, so I trust you'll upstream what you can!

PR #375, opened 2024-03-14
kas: version bump of base layers and build optimisations

TheYoctoJester:


MuchoLucho (approved):
Nothing out of the standard process.

PR #376, opened 2024-03-14
fix: kas: use poky as base distribution for all builds

TheYoctoJester:
Changelog: Title
Ticket: None


MuchoLucho (approved):
Everything looks normal

PR #386, opened 2024-07-10
Add meta-mender-client-only

TheYoctoJester:
Add the meta-mender-client-only layer, including example kas configuration and adding it to the automated build.

### DIFF
--- a/meta-mender-client-only/README
+++ b/meta-mender-client-only/README
@@ -1,0 +1,36 @@
+This README file contains information on the contents of the meta-mender-client-only layer.
+
+Please see the corresponding sections below for details.
+
+Dependencies
+============
+
+  URI: https://git.yoctoproject.org/poky
+  branch: scarthgap
+
+  URI: https://github.com/mendersoftware/meta-mender
+  branch: scarthgap
+
+Contributions
+=============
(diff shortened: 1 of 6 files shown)


meta-mender-client-only/README
```diff
@@ -0,0 +1,36 @@
...
+=================================================
+
+Run 'bitbake-layers add-layer meta-mender-client-only'
+
+II. Misc
+========
+
+This layer enables the Mender client to be installed and operated without the full integration. With the layer added to the build, you can add `mender-auth mender-update` and optionally `mender-connect` to the `IMAGE_INSTALL` variable and have those installed. The `MACHINE` will be used as device type.
+
+With client and connect installed, you can use all Update Modules and the Troubleshoot AddOn. Installing a root filesystem update is not supported.
```
lluiscampos: According to [our docs](https://docs.mender.io/add-ons),

```suggestion
With client and connect installed, you can use all Update Modules and the Troubleshoot Add-On. Installing a root filesystem update is not supported.
```

lluiscampos (approved):
Good initiative!

IMO the three commits could be squashed into a single one

PR #389, opened 2024-07-15
chore: kas: add beaglebone-uboot config

TheYoctoJester:
The BeagleBone Black can also use the mender-uboot feature instead of mender-grub. Add a build configuration to reflect that and autobuild it to catch eventual build failures.

Changelog: Title
Ticket: None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,8 @@ jobs:
           qemux86-64,
           vexpress-qemu,
           vexpress-qemu-flash,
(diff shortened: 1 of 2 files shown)


TheYoctoJester:
Thoughts @drewmoseley?

drewmoseley:
This sounds reasonable to me.

TheMeaningfulEngineer (approved):
An extra case to test the build for BBB with uboot only.
Haven't tested, but looks good.

PR #393, opened 2024-07-18
add the current OE4T state as base for upstreaming Orin support

TheYoctoJester:
- The Jetson Nano, TX2 boards are still supported on `kirkstone`, using Jetpack 4. There is no reason to not keep our support for it alive.
- Orin is the newer generation, and only supported on Jetpack 5.

Those are essentially incompatible, so a layer split was discussed and agreed upon with the https://github.com/OE4T community.

This series of patches moves the existing `meta-mender-tegra` layer into the `meta-mender-tegra/meta-mender-tegra-jetpack4` directory, and instatiates a new `meta-mender-tegra/meta-mender-tegra-jetpack5` layer to support the newer BSPs. This includes the first board integration "AGX Orin".


TheMeaningfulEngineer (approved):
LGTM
Didn't test it but the description is clear enough what it achieves.

PR #398, opened 2024-08-09
[scarthgap] add Jetson Tegra w/ Jetpack 5.x support

zach-welch-aquabyte:
This branch works well on our custom Orin NX carrier board running a snapshot of `scarthgap` (circa 5.0.2). It includes a few cherry-picked commits from Josef's branch.


TheYoctoJester (commented):
Hi @zach-welch-aquabyte,

Thanks a lot for assembling this. Two things should be changed, though:
- the meta-mender-tegra-jetpack4 layer is not used anymore, so it should be removed.
- with the JP4 support gone, the content of meta-mender-tegra-jetpack5 should be moved back to meta-mender-tegra.

Do you want to adjust this, or would like me to? I can prepare a PR on top of your branch.

zach-welch-aquabyte:
Thanks for the fast feedback. I will take care of these today. I’ll also
refine the commit messages’ metadata to improve the changelog.

zach-welch-aquabyte:
I have rebased the branch to pick the `meta-mender-tegra-jetpack5` sublayer from `kirkstone` directly into `meta-mender-tegra`. As promised, I also improved the metadata, so the ChangeLog looks more readable and complete.

During that review, I noticed that I unintentionally had squashed an unrelated change into an early commit. I have broken that out separately as the WIP patch near the end of the series. I have posted it here as-is for feedback about how to fix that last (known) issue.

TheYoctoJester:
Hi @zach-welch-aquabyte, great! I have already added my signed commits (required for me as part of Mender), and the rest looks good to me. Only one last nitpick is, e1017d7810ccde96ddf4efd0193a7eb67d777749 doesn't have a Signed-off-by. Once you add that its good to merge 👍

kas/jetson-agx-orin-devkit.yml
```diff
@@ -0,0 +1,11 @@
+header:
+  version: 14
+  includes:
+  - kas/include/mender-full.yml
+  - kas/include/tegra-jetpack6.yml
```
apbr, line 5: This file implies compatibility with JP 6. Do you know if this was ever tested?
For me Bitbake is complaining that it cannot apply `linux-tegra_%.bbappend` as the base recipe was renamed to `linux-jammy-nvidia-tegra`.

zach-welch-aquabyte:
Per @madisongh's suggestion, I resolved the pending issue with e1017d7810ccde96ddf4efd0193a7eb67d777749 in this PR by introducing the new global `l4t_version.bbclass` in `meta-tegra`, submitted as OE4T/meta-tegra#1658.

While that PR needs to be merged before this one, I have pushed my latest branch here to unblock that possibility.

meta-mender-tegra/conf/layer.conf
```diff
@@ -0,0 +1,12 @@
+# We have a conf and classes directory, add to BBPATH
+BBPATH .= ":${LAYERDIR}"
+# We have recipes-* directories, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
+		${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "meta-mender-tegra-jetpack5"
```
apbr: Shouldn't this be `meta-mender-tegra` now?
Same for the overrides that follow.

kas/include/tegra-jetpack5.yml
```diff
@@ -0,0 +1,19 @@
...
+  meta-tegra-community:
+    # refers to the scarthgap-l4t-r35.x branch
+    commit: 865561b6d3ba6607a814354c122c91422df720d5
+  meta-tegrademo:
+    # refers to the scarthgap-l4t-r35.x branch
+    commit: 2d9319ba0c24fb46937a742d78fc046983edb325
+
+  meta-mender-community:
+    layers:
+      meta-mender-tegra/meta-mender-tegra-jetpack5:
```
apbr: This kas file and the tegra-jetpack6 one probably also needs to be updated for the new layer location/name.

kas/jetson-agx-orin-devkit.yml
zach-welch-aquabyte: I have only tested with JP5. I suspect you might be the first to try it with JP6.

Does it build if you rename the `bbappend` file? While I am not certain how to resolve the renamed recipe issue correctly, it would be interesting to hear how much more work remains to be done, as I eventually plan to give JP6 a go as well.

meta-mender-tegra/conf/layer.conf
zach-welch-aquabyte: Good catch. I have posted a new patch that fixes this.

kas/include/tegra-jetpack5.yml
zach-welch-aquabyte: I have posted another patch that fixes this.

kas/jetson-agx-orin-devkit.yml
apbr: It builds, but there are still some other fixes needed to work properly. Though most of them seem to be compatible with JP 6.
I'm still in the process of figuring out all required changes :D

TheYoctoJester:
@apbr so we keep this open until sorted out? Or what strategy should we
apply?

Jonathan Schnitzler ***@***.***> schrieb am Fr. 16. Aug. 2024
um 08:46:

apbr:
> @apbr so we keep this open until sorted out? Or what strategy should we apply?

Good question, There are some incompatibilities I noticed that might be difficult to resolve (See Gitter/Matrix chat). I'm not sure how to proceed with those.
So not sure it makes sense to wait. Alternatively this could be merged as JP 5 only? What do you think?

TheYoctoJester:
JP5 only sounds like a good plan to me.

Jonathan Schnitzler ***@***.***> schrieb am Fr. 16. Aug. 2024
um 09:38:

meta-mender-tegra/recipes-bsp/tegra-binaries/redundant-boot-overrides_%.bbappend
```diff
@@ -0,0 +1 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
```
apbr: I just noticed this file appends `redundant-boot-overrides_1.0.bb` from the same layer.
To me it seems like this append and redundant-boot-overrides_1.0.bb could be just merged.
Do you understand why that is the case?

meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-nv-boot-control-config_%.bbappend
```diff
@@ -0,0 +1,3 @@
+do_install:append() {
+    ln -sf /run/nv_boot_control/nv_boot_control.conf ${D}${sysconfdir}/
```
apbr: Any idea if this is still needed?
This already seems to be part of meta-tegra since quite some time (e.g. https://github.com/OE4T/meta-tegra/blob/scarthgap-l4t-r35.x/recipes-bsp/tegra-binaries/tegra-nv-boot-control-config_35.5.0.bb#L23)

meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-redundant-boot_%.bbappend
```diff
@@ -0,0 +1,16 @@
+EXTRADEPS = "redundant-boot-overrides"
```
apbr: Afaik we only support `tegra234`.
So `redundant-boot-overrides` will never be used, right?
Do you see any reason why we shouldn't delete it and it's corresponding files?
Or am I missing something?

apbr (commented):
I added some more comments.
(I know, you probably didn't write most of the changes in question,)

meta-mender-tegra/recipes-bsp/tegra-binaries/redundant-boot-overrides_%.bbappend
zach-welch-aquabyte: I noticed that as well. No idea why they are separate. Unless someone has a good explanation (and there very well might be one), I agree that it looks like they should be merged.

meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-redundant-boot_%.bbappend
zach-welch-aquabyte: Another good catch. I have pushed a new patch to drop this recipe.

meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-nv-boot-control-config_%.bbappend
zach-welch-aquabyte: Good catch. I pushed a patch to drop this bbappend.

meta-mender-tegra/recipes-bsp/tegra-binaries/redundant-boot-overrides_%.bbappend
zach-welch-aquabyte: After further review of the broader context, this issue was resolved by nuking the entire recipe from high orbit.

kas/jetson-agx-orin-devkit.yml
TheYoctoJester: @apbr I'll go ahead and merge it as-is then, to reduce the amount of things which need to be carried along for JP6. To me its nicer having a state in-branch which already builds and "just" needs some more polish, instead of having to keep a whole separate state externally for even working on JP6.

kas/demos/raspberrypi4-64-app-updates.yml
```diff
@@ -1,5 +1,5 @@
 header:
-  version: 11
+  version: 14
```
TheYoctoJester: should be removed from this branch

kas/demos/qemuarm64-client-only.yml
```diff
@@ -1,5 +1,5 @@
 header:
-  version: 11
+  version: 14
```
TheYoctoJester: should be removed from this branch

kas/include/mender-base.yml
```diff
@@ -20,7 +20,7 @@ repos:
 
   meta-mender:
     url: https://github.com/mendersoftware/meta-mender.git
-    commit: 4302cb71a7f8a0ffb09f47e41f4ce1df3e812419
+    commit: 21b8bd76e92391665ad2cf52fa1def65ba97c7a2
```
TheYoctoJester: needs bump to e8b6b3e554392f6d0af3eb18667d93e658205250 so your PR on persistent machine IDs is already merged

kas/include/tegra-base.yml
```diff
@@ -0,0 +1,42 @@
+header:
+  version: 14
+
+repos:
+  meta-mender:
```
TheYoctoJester: meta-mender can be removed then here

meta-mender-tegra/README.md
```diff
@@ -0,0 +1,51 @@
...
+- AGX Xavier
+- Orin Nano
+
+
+## Dependencies
+
+This layer depends on:
+
+```
+URI: https://github.com/madisongh/meta-tegra.git
```
TheYoctoJester: should be https://github.com/OE4T/meta-tegra, I think?

meta-mender-tegra/README.md
```diff
@@ -0,0 +1,51 @@
...
+layers: meta-tegra
+branch: kirkstone
```
TheYoctoJester: should be `scarthgap`, respectively `scarthgap-l4t-r35.x`

meta-mender-tegra/README.md
```diff
@@ -0,0 +1,51 @@
...
+revision: HEAD
+```
+
+```
+URI: https://github.com/mendersoftware/meta-mender.git
+layers: meta-mender-core
+branch: kirkstone
```
TheYoctoJester: should be `scarthgap`

TheYoctoJester (changes requested):
Hi @zach-welch-aquabyte, gave it a review on mostly the formal things I can reasonably comment on. I can also fix those and push, if you'd like me to.

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/switch-rootfs
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+set -e
+echo "$(mender show-artifact): $(basename "$0") was called!" >&2
```
apbr, line 3: @zach-welch-aquabyte Maybe a bit unrelated, but on my device/image the binary is only available as `mender-update` and not as `mender`. Do you have any idea, why that could be the case?

TheYoctoJester: @apbr starting in Mender Client 4.0, the binary is split into `mender-auth` and `mender-update`. So unless additional configuration is applied, this script should actually refer to `mender-update`.

apbr: Ok, I proposed a change: https://github.com/zach-welch-aquabyte/meta-mender-community/pull/1 (Includes also a first step towards compatibility with JP 6)

TheYoctoJester (approved):

meta-mender-tegra/recipes-mender/tegra-state-scripts/files/switch-rootfs
apbr: As this PR is now merged, I created a PR directly here: https://github.com/mendersoftware/meta-mender-community/pull/402

PR #402, opened 2024-08-21
Scarthgap tegra jetpack5 v3

apbr:
Starting with Mender Client 4.0, the binary is split into mender-auth and mender-update, so we should use `mender-update`.
Also includes a first step towards compatibility with JP 6.

### DIFF
--- a/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/abort-blupdate
+++ b/meta-mender-tegra/recipes-mender/tegra-state-scripts/files/abort-blupdate
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-CAPTARGET="/opt/nvidia/esp/EFI/UpdateCapsule/TEGRA_BL.Cap"
+CAPTARGET="/boot/efi/EFI/UpdateCapsule/TEGRA_BL.Cap"
 
 if [ -f "$CAPTARGET" ]; then
     rm "$CAPTARGET"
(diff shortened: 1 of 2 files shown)


zach-welch-aquabyte (changes requested):
I have been testing with the 3.x client and only recently tried to switch to 4.x, as the fixes required for it to build for us only recently landed upstream. My initial testing with 4.x on this branch was... unsuccessful. I ended up rolling back to 3.x for the time being, in part because I don't want to change workflows that use the old CLI syntax. I will be investigating further, but I need to use 3.x in the meantime.

Naively, I might propose a simple `mender` wrapper script be installed by the 4.x recipe in `meta-mender-core`. This would call through to the new 4.x commands, allowing shell-driven workflows written for 3.x to continue working as expected with either version. This might need to be optional, enabled only via `PACKAGECONFIG[mender-3-compat]` or similar. In writing this, I imagine there are others that have their own automation built on the mender CLI, so I wonder if there might be a good reason that Mender hasn't added this already.

If that change isn't feasible for the core, then the `switch-rootfs` script might instead be updated to detect which command is available and do the right thing accordingly. Compatibility achieved.

apbr:
@zach-welch-aquabyte Oh, I assumed the Mender client pre 4.0 could be called by `mender` or `mender-update`. Is that not the case?

TheYoctoJester:
@apbr back from AFKness now, sorry for the delay. Is this still relevant? My understanding is that it "just" fixes the EFI mount paths. As I can't judge, can @zach-welch-aquabyte or @dwalkes give me a yay or nay please?

apbr:
> My understanding is that it "just" fixes the EFI mount paths.

That is a requirement for mender updates to work on JP6. (The old EFI mount path is not available anymore on JP6)
Though as the other part of this PR (calling `mender-update` instead of `mender`) is not general enough, this PR probably shouldn't be merged.

PR #406, opened 2024-10-11
feat: add meta-mender-validation

TheYoctoJester:


meta-mender-validation/recipes-mender/bootloader-validation/bootloader-validation_0.1.bb
```diff
@@ -0,0 +1,31 @@
...
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=86d3f3a95c324c9479bd8986968f4327"
+
+SRC_URI = "git://github.com/theyoctojester/mender-validation.git;protocol=https;branch=main"
+SRCREV = "8e9ed9f22b1a7e193dbb0ce93c27ff2b2d9e804b"
+
+do_configure[noexec] = "1"
+do_compile[noexec] = "1"
+
+RDEPENDS:${PN} = "python3"
```
lluiscampos, line 15: We could, maybe, fine tune a bit the dependency with:
```suggestion
RDEPENDS:${PN} = "python3-core python3-json"
```

Although it is easy to miss packages this way

lluiscampos (approved):
It looks good!

meta-mender-validation/recipes-mender/bootloader-validation/bootloader-validation_0.1.bb
TheYoctoJester: @lluiscampos yes I have considered that, but decided that the validation script (including those to come) are allowed to use the full python standard library unconditionally, and nothing else.

So I'd like to keep it like that for the future, e.g. eventual extensions will not break because of the recipe being restrictive.

I'll add a short explanation comment though, it's a really good catch!🙏

PR #408, opened 2024-10-23
chore: bump layer revisions for kirkstone

TheYoctoJester:


kas/include/tegra-jetpack4.yml
```diff
@@ -1,20 +1,23 @@
 header:
-  version: 11
+  version: 14
   includes:
   - kas/include/tegra-base.yml
 
 repos:
   meta-tegra:
-    refspec: kirkstone-l4t-r32.7.x
+    #branch: kirkstone-l4t-r32.7.x
```
oldgiova, line 8: Is this supposed to be commented?

oldgiova (approved):
Just a comment, but LGTM for what I could check

kas/include/tegra-jetpack4.yml
TheYoctoJester: @oldgiova thanks, that is actually intended. The concept is: "when the commit SHA refers to a branch which ist *not* the canonical release one, a comment shows the used one". So the canonical release branch is `kirkstone`, and almost all layers adhere to it, just this for example needs the `kirkstone-l4t-r32.7.x` branch to build.

PR #417, opened 2024-11-08
chore: used fixed revisions for NXP layers

TheYoctoJester:
Changelog: Title
Ticket: None


oldgiova (approved):
Approving. Maybe is better to create a named branch out of this commit?

PR #422, opened 2024-11-18
fix: u-boot on Raspberry Pi for 32 bit mode

TheYoctoJester:


meta-mender-raspberrypi/recipes-bsp/u-boot/u-boot-raspberrypi.inc
```diff
@@ -6,9 +6,12 @@ BOOTENV_SIZE:rpi ?= "0x4000"
 # use of boot.scr and cmdline.txt.
 MENDER_UBOOT_AUTO_CONFIGURE:rpi = "0"
 
+SRC_URI:append:rpi = " file://memory_size.cfg"
```
lluiscampos, line 9: Isn't this configuration uboot specific?
```suggestion
SRC_URI:append:rpi:mender-uboot = " file://memory_size.cfg"
```

TheYoctoJester: @lluiscampos it is `u-boot` specific, hence in the `u-boot` recipe, but its not Mender specific. So not tied to the `mender-uboot` feature.

Technically both of those *could* go to `meta-raspberrypi`, but I want to unblock our users first, then care about upstream.

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-pci-fdt-ignore-checks-for-address-range.patch
ghost, line 1: > Commit 78126711cb2167931775cff84acfcba17b07689f in u-boot upstream
> adds "pci: Check region ranges are addressable". This blocks the
> booting process on Raspberry Pi platforms running in 32bit mode.

I can't find the commit that this paragraph is talking about. Which repo is it?

ghost (approved):
I can't really say much about the validity of the address patch, but seems sensible. 🙂

PR #427, opened 2024-12-02
tegra: fixes for Jetpack6

TheYoctoJester:


poett1:
Just tried `jetson-agx-orin-devkit-64`, which builds and flashes fine without changes. However, pushing an update using mender-demo fails with:

```
2024-12-02 16:56:29.723 +0000 UTC info: Running Mender client 4.0.4
2024-12-02 16:56:29.724 +0000 UTC info: Deployment with ID 66927674-865f-4132-8e83-31a781d944ad started.
2024-12-02 16:56:29.724 +0000 UTC info: Sending status update to server
2024-12-02 16:56:29.845 +0000 UTC info: Installing artifact...
2024-12-02 16:56:29.905 +0000 UTC info: Update Module output (stderr): ERR: could not identify current boot slot
2024-12-02 16:56:29.906 +0000 UTC info: Update Module output (stderr): /usr/share/mender/modules/v3/rootfs-image: line 96: test: UNKNOWN: integer expression expected
2024-12-02 16:56:29.911 +0000 UTC info: Update Module output (stderr): Mounted root does not match boot loader environment (/dev/mmcblk0p2)!
2024-12-02 16:56:29.911 +0000 UTC error: Process returned non-zero exit status: Download: Update Module returned non-zero status: Process exited with status 1
2024-12-02 16:56:29.921 +0000 UTC info: Sending status update to server
```

It seems like `core-image-minimal` does not include the necessary tools (e.g. `nvbootctrl / tegra-redundant-boot` is not installed). I changed the target image to `demo-image-base` to include the default Tegra stuff for testing and it seems like the update was successful: 
```
root@p3737-0000-p3701-0005:~# nvbootctrl dump-slots-info
Current version: 36.3.0
Capsule update status: 1
Current bootloader slot: B
Active bootloader slot: B
num_slots: 2
slot: 0,             status: normal
slot: 1,             status: normal
```

@TheYoctoJester We could use `demo-image-base` as default, but this will result in much more packages to build compared to `core-image-minimal`, which is pretty bad for CI workflows. Should we add the necessary packages to the core-image-minimal build instead? Is there an official (automated) test procedure for Mender OTA to ensure reliability of the client implementation?

Do we need to test all supported Orin devices before merging? I only have AGX Orin Developer Kit 64GB and Orin NX 16GB (on Xavier NX Developer Kit carrier) at hand, so other people need to try this on Orin Nano.

In addition, this PR covers L4T 36.3/JP6.0 commits right now, latest OE4T scarthgap has already been updated to L4T 36.4/JP6.1. L4T 36.4/JP6.1 may require some additional work. We might also consider adding some sort of versioning here to make the update process from L4T36.3-->L4T36.4 easier?

TheYoctoJester:
Hi @poett1 ,

> It seems like `core-image-minimal` does not include the necessary tools (e.g. `nvbootctrl / tegra-redundant-boot` is not installed). I changed the target image to `demo-image-base` to include the default Tegra stuff for testing and it seems like the update was successful </snip>
> @TheYoctoJester We could use `demo-image-base` as default, but this will result in much more packages to build compared to `core-image-minimal`, which is pretty bad for CI workflows. Should we add the necessary packages to the core-image-minimal build instead? Is there an official (automated) test procedure for Mender OTA to ensure reliability of the client implementation?

The easiest and most straightforward solution is to just have `tegra-state-scripts` set an `RDEPENDS` on `tegra-redundant-boot`, because then it will work for whatever image. Just pushed the patch.👍

> Do we need to test all supported Orin devices before merging? I only have AGX Orin Developer Kit 64GB and Orin NX 16GB (on Xavier NX Developer Kit carrier) at hand, so other people need to try this on Orin Nano.

Definitely not. Orin Nano is not even building correctly at the moment, and I'm working on figuring out why. So merging this will not cause any harm.

> In addition, this PR covers L4T 36.3/JP6.0 commits right now, latest OE4T scarthgap has already been updated to L4T 36.4/JP6.1. L4T 36.4/JP6.1 may require some additional work. We might also consider adding some sort of versioning here to make the update process from L4T36.3-->L4T36.4 easier?

Good question. Do you happen to know which changes that would be? Is there anything in this layer that is actually version dependent?

Austriker:
@TheYoctoJester 

I had a go trying to build the Jetson Orin NX with Jetpack 6 on your branch.

The KAS config file:
```yml
header:
  version: 14
  includes:
  - kas/include/mender-full.yml
  - kas/include/tegra-jetpack6.yml

machine: p3768-0000-p3767-0000

local_conf_header:
  AB-upgrades: |
    USE_REDUNDANT_FLASH_LAYOUT_DEFAULT = "1"
```

But I get the following error:
```
ERROR: Failure expanding variable MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT:tegra, expression was ${@tegra_mender_calc_total_size(d)} which triggered exception ValueError: invalid literal for int() with base 10: ''
The variable dependency chain for the failure is: MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT:tegra
```

What is the best way to solve this issue ? 

I had a look at this thread: [Plans or Methods for Orin NX Support in Yocto?](https://hub.mender.io/t/plans-or-methods-for-orin-nx-support-in-yocto/7211/9) but the branch you created with the specific changes you made doesn't exist anymore.

TheYoctoJester:
@Austriker 

> ```
> ERROR: Failure expanding variable MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT:tegra, expression was ${@tegra_mender_calc_total_size(d)} which triggered exception ValueError: invalid literal for int() with base 10: ''
> The variable dependency chain for the failure is: MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT:tegra
> ```
> 
> What is the best way to solve this issue ?
> 
> I had a look at this thread: [Plans or Methods for Orin NX Support in Yocto?](https://hub.mender.io/t/plans-or-methods-for-orin-nx-support-in-yocto/7211/9) but the branch you created with the specific changes you made doesn't exist anymore.

The error message is really misleading. Just worked through it, and added a patch with a proper error message. You need to set `EMMC_SIZE`, have also added 32GiB as default for `jetson-orin-nano-devkit.yml`. Thoughts?

Austriker:
@TheYoctoJester I was also working this issue. Thank you for your update.

Would you also set `EMMC_SIZE` in case of NVME only like for Nvidia Jetson Orin NX ?

I haven't flashed the build yet. I was going for something like this in the local.conf
```
# Workarounds to avoid the following error:
# The variable dependency chain for the failure is: MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT:tegra
# https://hub.mender.io/t/plans-or-methods-for-orin-nx-support-in-yocto/7211/5?u=austriker
# https://hub.mender.io/t/support-of-orin-nano-devkit-nvme-on-jetpack-5-0/7087/2
# $ fdisk -l /dev/nvme0n1
# Disk /dev/nvme0n1: 447.13 GiB, 480103981056 bytes, 937703088 sectors
EMMC_SIZE = "0"
MENDER_STORAGE_TOTAL_SIZE_MB = "457862"
TEGRA_EXTERNAL_DEVICE_SECTORS = "937703088"
```

Or based on your change just set:
``` 
EMMC_SIZE = "479 962 595 328"
```

what are your thoughts ?

TheYoctoJester:
@Austriker 

> Would you also set `EMMC_SIZE` in case of NVME only like for Nvidia Jetson Orin NX ?

Probably, but I don't have experience with that. @poett1 can probably comment.

> I haven't flashed the build yet. I was going for something like this in the local.conf
> 
> ```
> # Workarounds to avoid the following error:
> # The variable dependency chain for the failure is: MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT:tegra
> # https://hub.mender.io/t/plans-or-methods-for-orin-nx-support-in-yocto/7211/5?u=austriker
> # https://hub.mender.io/t/support-of-orin-nano-devkit-nvme-on-jetpack-5-0/7087/2
> # $ fdisk -l /dev/nvme0n1
> # Disk /dev/nvme0n1: 447.13 GiB, 480103981056 bytes, 937703088 sectors
> EMMC_SIZE = "0"
> MENDER_STORAGE_TOTAL_SIZE_MB = "457862"
> TEGRA_EXTERNAL_DEVICE_SECTORS = "937703088"
> ```
> 
> Or based on your change just set:
> 
> ```
> EMMC_SIZE = "479 962 595 328"
> ```
> 
> what are your thoughts ?

Using such a huge `EMMC_SIZE` is technically okay, but will lead to huge images and long flashing times. A better approach is probably to carefully choose the A/B partition sizes, and dynamically grow the data partition on first boot.

Austriker:
@TheYoctoJester ok so you would just set `EMMC_SIZE` to 32 GB as an example or do you need to fine tune the partition structure using variables like `MENDER_DATA_PART_SIZE_MB` ?

poett1:
> Probably, but I don't have experience with that. @poett1 can probably comment.

Me neither. 

@TheYoctoJester I fetched your latest changes, but `nvbootctrl `is still missing in `core-image-minimal`. I'm rebuilding and checking if it exists in rootfs:
```
kas build kas/jetson-agx-orin-devkit-64.yml
find <build_dir>/tmp/work/p3737_0000_p3701_0005-oe4t-linux/core-image-minimal/ -name nvbootctrl

```
This works fine to install it
```
local_conf_header:
  Deps: |
    IMAGE_INSTALL:append = " tegra-redundant-boot" 
```
but I can't tell why RDEPENDS does not work.

TheYoctoJester:
> but I can't tell why RDEPENDS does not work.

Hi @poett1 I just figured it out. The `tegra-state-scripts` package is actually not installed to the image, but packaged into the artifact, so the `RDEPENDS` never applied. Moving it to the Mender Client `EXTRADEPS` does the trick, thanks for checking!

tranchitella (approved):

PR #429, opened 2024-12-18
Add support for Jetson Orin NX

Austriker:
Enable support for a jetson Orin NX with NVMe

Based on this issue: https://hub.mender.io/t/jetson-orin-nx-and-jetpack-6-mender-intergration-issues/7375

- Create custom partition layout
- Use `<allocation_attribute> 0x808 </allocation_attribute>` to expand the user partition automatically during the flash process
- Add documentation

### DIFF
--- a/kas/jetson-orin-16gb-nx-p3786.yml
+++ b/kas/jetson-orin-16gb-nx-p3786.yml
@@ -1,0 +1,17 @@
+header:
+  version: 14
+  includes:
+    - kas/include/mender-full.yml
+    - kas/include/tegra-jetpack6.yml
+
+machine: p3768-0000-p3767-0000
+
+local_conf_header:
+  AB-upgrades: |
+    USE_REDUNDANT_FLASH_LAYOUT_DEFAULT = "1"
+  eMMC: |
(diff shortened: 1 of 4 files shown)


TheYoctoJester:
Hi @Austriker,

As you mentioned on IRC, I think changing to
```
PARTITION_FILE_EXTERNAL:p3768-0000-p3767-0000 = "${WORKDIR}/flash_l4t_t234_nvme_rootfs_ab.xml"
```
is a good approach. However it has the small shortcoming being tied to exactly this machine. So how about adding a short note to the layers README which explains why the line is there, and hints at adapting it for custom Orin NX based machines?

Austriker:
Hi @TheYoctoJester ,

I fixed the PR and added some documentation. I also remove the quick start guide since for me mender integration is not maintained anymore in tegrademo-distro.

TheYoctoJester:
Hi @Austriker,

I folded this PR into the bigger Tegra refactoring at https://github.com/mendersoftware/meta-mender-community/pull/439.
Can you check if this matches expectations, so we can eventually close here?

PR #431, opened 2025-01-10
chore: split the GitHub Actions build workflows

TheYoctoJester:
So far the build workflow was monolithic, which does not scale for a fast turnaround. To improve this, split into
- preparation stage, which caches all required metadata layers
- building the known good boards
- building the experimental boards
- building the boards which are also in automated validation.
The latter upon every push.

Validation build images contain the boot loader integration test script, and are collected to an output directory on NFS for being picked up by the test runners.

Changelog: Title
Ticket: None

### DIFF
--- a/ci/collect_artifacts.sh
+++ b/ci/collect_artifacts.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+
+if [ -z ${BUILDDIR+x} ]; then
+	echo "BUILDDIR is unset, exiting";
+	exit 1;
+fi
+if [ -z ${ARTIFACT_PATH+x} ]; then
+	echo "ARTIFACT_PATH is unset, exiting";
+	exit 1;
+fi
+
+echo "using BUILDDIR: $BUILDDIR"
+echo "using ARTIFACT_PATH: $ARTIFACT_PATH"
+
+mkdir -p $ARTIFACT_PATH
+
+IMAGE_SUBPATH="build/tmp/deploy/images"
+
+for build in $(ls $BUILDDIR); do
+	echo "found build: $build"
+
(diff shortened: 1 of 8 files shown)


ci/collect_artifacts.sh
```diff
@@ -27,13 +27,13 @@ for build in $(ls $BUILDDIR); do
...
-		cp -fR $BUILDDIR/$build/$IMAGE_SUBPATH $ARTIFACT_PATH/$board
-		cp -fR $BUILDDIR/$build/build/conf $ARTIFACT_PATH/$board
+		mkdir -p $ARTIFACT_PATH/$build
+		echo "$board" > $ARTIFACT_PATH/$build/board.txt
+		cp -fR $BUILDDIR/$build/$IMAGE_SUBPATH $ARTIFACT_PATH/$build
+		cp -fR $BUILDDIR/$build/build/conf $ARTIFACT_PATH/$build
 
 		if [ -d "$BUILDDIR/$build/build/tmp/buildstats" ]; then
-			cp -fR $BUILDDIR/$build/build/tmp/buildstats $ARTIFACT_PATH/$board
+			cp -fR $BUILDDIR/$build/build/tmp/buildstats $ARTIFACT_PATH/$build
```
lluiscampos: All this really belongs to a previous commit

lluiscampos (approved):
It looks good!

I dropped a comment but feel free to ignore if you don't want to modify the git history - I don't think it is very important in this repository.

ci/collect_artifacts.sh
TheYoctoJester: Thanks, cleaned it up!

PR #428, opened 2024-12-06
Remove the need to copy ownership for FAT partition

tactmaster:
Error on our system:
cp: failed to preserve ownership for '/uboot/start_cd.elf.tmp': Operation not permitted

* uboot parition is FAT on our system and rasbbery  PI OS. 
 Has this change in newer releases? 
 i.e. /dev/mmcblk0p1 on /uboot type vfat

PR #441, opened 2025-03-27
rpi-update-firmware.bbclass: refactor rpi_install_firmware_to_rootfs

TheYoctoJester:
The current rpi_install_firmware_to_rootfs has several problems:
- it's installing duplicated devicetrees to /boot/firmware, because it is installing all dtb/dtbo links but they are duplicated, for instance: uart2.dtbo and uart2-raspberrypi4-64.dtbo are pointing to a same file.
- While RPI bootfs contains all files listed in IMAGE_BOOT_FILES, a end user can customize that variable to install the files they want to be in boot partition, but rpi_install_firmware_to_rootfs does not handle it.

To fix the above, introduce a function mender_install_deployed_files which is implemented in the same way with how WIC handle IMAGE_BOOT_FILES, now the files in boot partition and the files in /boot/firmware of A/B partitions are consistent.

This is a direct port to meta-mender-community from https://github.com/mendersoftware/meta-mender/pull/2114/commits/86e0949a146efff718548ee8e259941c16b56c4f

### DIFF
--- a/meta-mender-raspberrypi/classes/rpi-update-firmware.bbclass
+++ b/meta-mender-raspberrypi/classes/rpi-update-firmware.bbclass
@@ -1,15 +1,63 @@
-rpi_install_firmware_to_rootfs() {
-    install -d ${IMAGE_ROOTFS}/boot/firmware/overlays
+# Whitespace separated list of files declared by 'deploy_var' variable
+# from 'source_dir' (DEPLOY_DIR_IMAGE by default) to place in 'deploy_dir'.
+# Entries will be installed under a same name as the source file. To change
+# the destination file name, pass a desired name after a semicolon
+# (eg. u-boot.img;uboot). Exactly same rules with how IMAGE_BOOT_FILES being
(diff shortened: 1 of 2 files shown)


TheYoctoJester:
@tactmaster @liuming50 please comment/review here.

tactmaster:
Other than my comments on https://github.com/mendersoftware/meta-mender-community/pull/428 this looks fine.

I haven't tested it directltly. if you would like me to I will

tactmaster (approved):

TheYoctoJester:
Added over at https://github.com/mendersoftware/meta-mender-community/pull/441, thoughts?

michaek:
@TheYoctoJester Was this meant to be stale/closed? It seems like your last comment was meant to be somewhere else drawing attention to this, as it currently just links to itself. I'd like to see the rpi-update-firmware task not include the Image file, as it tends to fill up the boot partition needlessly, and it seems like this would do the trick. Thanks!

PR #442, opened 2025-04-08
meta-mender-tegra: remove u-boot

pberginkonsult:
Since Jetpack 5 the default bootloader is UEFI based on edk2. meta-tegra layer used here have removed all stuff related to u-boot this commit  clean up meta-mender-tegra.


TheYoctoJester (approved):

cakre:
This removed `libubootenv-fake` which was used to set and read the `upgrade_available` flag causing `mender-commit`, `nv_update_verifier`etc. to break. Because of this updates fail and are rolled back in every case (and the install breaks completely after two reboots after flashing)

TheYoctoJester:
> This removed `libubootenv-fake` which was used to set and read the `upgrade_available` flag causing `mender-commit`, `nv_update_verifier`etc. to break. Because of this updates fail and are rolled back in every case (and the install breaks completely after two reboots after flashing)

Ouch! Sorry to hear - unfortunately I can't (yet) test the Nvidia integrations so I have to rely on PRs. How should we proceed @cakre @pberginkonsult ? Revert? Or a follow-up PR to re-add?

pberginkonsult:
> > This removed `libubootenv-fake` which was used to set and read the `upgrade_available` flag causing `mender-commit`, `nv_update_verifier`etc. to break. Because of this updates fail and are rolled back in every case (and the install breaks completely after two reboots after flashing)
> 
> Ouch! Sorry to hear - unfortunately I can't (yet) test the Nvidia integrations so I have to rely on PRs. How should we proceed @cakre @pberginkonsult ? Revert? Or a follow-up PR to re-add?

Sorry for the mess! Ideally I think it seems wrong to me to rely on libubootenv and creating a fake package for that when not using u-boot. But can not look in to this immediately. So I think the quickest path is to just revert this.

cakre:
I think reverting would be quickest and easiest

cakre:
> > > This removed `libubootenv-fake` which was used to set and read the `upgrade_available` flag causing `mender-commit`, `nv_update_verifier`etc. to break. Because of this updates fail and are rolled back in every case (and the install breaks completely after two reboots after flashing)
> > 
> > 
> > Ouch! Sorry to hear - unfortunately I can't (yet) test the Nvidia integrations so I have to rely on PRs. How should we proceed @cakre @pberginkonsult ? Revert? Or a follow-up PR to re-add?
> 
> Sorry for the mess! Ideally I think it seems wrong to me to rely on libubootenv and creating a fake package for that when not using u-boot. But can not look in to this immediately. So I think the quickest path is to just revert this.

The libubootenv stuff is not for uboot in jetpack directly but to wrap arount menders uboot support. We basically immitate uboot for mender to be happy and rely nvidias boot chain switching logic internally (using the retry counter and `nvbootctrl verify` depending on if we are currently updating/committing)

TheYoctoJester:
@pberginkonsult @cakre can you comment and review over at https://github.com/mendersoftware/meta-mender-community/pull/445 so it can be merged?

On the whole topic complex, the technically sound way would actually be to just fold the whole logic into a custom `roots-image` Update Module. Up to Mender Client 3.x, the default one was built in and could not be easily overridden. Since Mender Client 4.x, it is factored out into https://github.com/mendersoftware/mender/blob/master/support/modules/rootfs-image, so it should be quite easy to completely remove the `fw_getenv/fw_setenv` logic and wrappers, and insert the Tegra specific logic directly.

PR #444, opened 2025-04-23
variscite: add support for imx8mp-var-dart

yairbenavraham:


TheYoctoJester:
Hi @yairbenavraham,

I have rebased the branch tried to build it, but so far it fails with
```
ERROR: imx-boot-1.0-r0 do_compile: ExecutionError('/home/mender/mmc-upstreaming/variscite/meta-mender-community/my-imx8mp-var-dart/build/tmp/work/imx8mp_var_dart-fslc-linux/imx-boot/1.0-r0/temp/run.do_compile.1927036', 1, None, None)
```

My guess is that there is a layer revision discrepancy somewhere, can you please provide the banner list of layers and their revisions which `bitbake` present when starting the build, and the additions to a blank `local.conf` this requires?

Thanks!

PR #451, opened 2025-05-03
auto grow UDA

rishabnayak:
This describes a mechanism to auto grow UDA, it is *not* reserved by Nvidia contrary to their response on the discussion forum/has changed in recent jetpack releases.

This has been tested to work on Xavier NX.

### DIFF
--- a/meta-mender-tegra/README.md
+++ b/meta-mender-tegra/README.md
@@ -30,16 +30,16 @@ revision: HEAD
 ## Layer structure
 
 - `meta-mender-tegra-common`
-Holds all common parts of the Mender integration for Tegra across all
-currently supported releases of Jetpack
+  Holds all common parts of the Mender integration for Tegra across all
+  currently supported releases of Jetpack
 
 - `meta-mender-tegra-jetpack5`
-Holds Jetpack release 5 specific parts of the Mender integration for Tegra.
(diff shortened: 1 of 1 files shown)


TheYoctoJester:
@cakre thoughts here maybe?

cakre:
> @cakre thoughts here maybe?

This works. I do this on my orin nano as well with a custom partition layout xml. It's particularly useful on nvme. I give it id=16 though (tbf i haven't checked against the latest stock flash layout. Maybe I just lack an optional partition on my end).

I can't really say anything about the example since I'm not that fond of awk syntax. 
With this generic solution though maybe this could be put directly into meta-mender-tegra and made configurable through a variable? I have used a setup like this on Jetson Nano, Xavier NX and Orin Nano so far so this grow flag is pretty universal. Just an idea though

TheYoctoJester (commented):
Hi @rishabnayak,

Thanks for the submission. Two things to change/consider:
- in order to merge, the patch needs to be signed off (`git commit -s`)
- would it make sense to not replace the `README` section, but add a second one so users can choose?

Greetz,
Josef

rishabnayak:
Hey @TheYoctoJester, resolved both!

TheYoctoJester (approved):

PR #454, opened 2025-05-13
tegra: Detect powerloss during update using boot_attempt flag

cakre:
On tegra devices the update capsule process does not catch powerlosses turing boot of the new rootfs and simply keeps on trying to boot it. Since rerunning several state scripts (e.g. Reboot_Leave -> Reboot_Leave) we could end up in an undefined state.   
To mitigate this `nv_update_verifier` now remembers if we already tried booting the current rootfs in an upgrade process and aborts/reboots if necessary.  
Note that this does not catch power losses that happen before the `nv_update_verifier` service starts and by that extend also does not catch power losses before systemd starts.

### DIFF
--- a/meta-mender-tegra/meta-mender-tegra-common/recipes-bsp/tegra-binaries/tegra-redundant-boot/nv_update_verifier.sh
+++ b/meta-mender-tegra/meta-mender-tegra-common/recipes-bsp/tegra-binaries/tegra-redundant-boot/nv_update_verifier.sh
@@ -1,9 +1,25 @@
 #!/bin/sh
 
+boot_flag="/var/lib/mender/boot_attempted"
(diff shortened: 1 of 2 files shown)


cakre:
I just had an idea to make the reboot a bit faster when force failing the new chain. im gonna push that real quick if that's fine with you? Including a rebase of course

TheYoctoJester (approved):

TheYoctoJester:
@cakre fun coincidence, I was just about to merge. Sure, fire right away. Rebase is already done.

cakre:
Ok, I just updated it to use `systemctl reboot -f` instead of `reboot`. Its ready to go. Thanks for waiting ^^

PR #456, opened 2025-05-22
kas: demos: fix qemuarm64-clientonly

TheYoctoJester:
Packaging mender-connect depends on the mender-systemd overrride being set. Without mender-full being inherited in the build, this is not populated into DISTROOVERRIDES.

As a quick fix, add it via local.conf.


elkoniu (approved):
ACK

PR #463, opened 2025-07-02
kas: migrate to kas >= 4.8 and security updates

aduskett:
Kas 4.8 introduced some compelling new security features that are quire useful for verification purposes. As mender prides itself on security, now would be a great time to update to kas 4.8 and add signature support where possible.

  - Update all .yml files to version 19.

  - Add YoctoBuildandRelease, MetaArm, and NorthernTech signers in the kas/include/mender-base.yml file.

  - Add tags where possible for all repositories.

  - Add branches where possible for all repositories. Note: If a tag exists, branches may not be used.

  - Update commits to the latest tag or commit for all scarthgap branches.

  - Add `signed: true` wherever possible.

  - Even though `signed: false` is the default, it is best to explicitly add it to every repository that doesn't have a signed tag. This lets the user know that the commit isn't signed and leaves no room for ambiguity.

### DIFF
--- a/kas/include/sulka.yml
+++ b/kas/include/sulka.yml
@@ -6,7 +6,7 @@ distro: sulka
 repos:
   meta-sulka-distro:
     url: https://codeberg.org/AltidSec/meta-sulka-distro.git
-    commit: 45c32dcd77ab35f91dca56a5265bdeaeb74f9766
+    commit: 64c1f5c0d484bd201e7f2f617040e2c5973d17cb
 
   meta-openembedded:
     layers:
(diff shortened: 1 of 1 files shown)


TheYoctoJester:
Hi @aduskett,

Thanks a lot for this PR. I've had looking at `kas`' signing mechanism on my laundry list for a long time now, seems I need to really get started. So it's great that you kicked me, but this one will definitely take some time as I really need to understand the implications here, especially with regards to who signs what, and in which authority. I'll keep you posted!

aduskett:
> Hi @aduskett,
> 
> Thanks a lot for this PR. I've had looking at `kas`' signing mechanism on my laundry list for a long time now, seems I need to really get started. So it's great that you kicked me, but this one will definitely take some time as I really need to understand the implications here, especially with regards to who signs what, and in which authority. I'll keep you posted!

No worries! I appreciate the review and keeping me in check! :D

PR #477, opened 2025-08-16
scarthgap: add jetpack 6 support for jetson-orin-nano-devkit-nvme

RoopanJK:
1. Add Jetpack 6 support for board `jetson-orin-nano-devkit-nvme`.
2. Fix typo in partition 17's name in layout file `flash_l4t_t234_nvme_rootfs_ab.xml`


RoopanJK:
I'm currently working on adding Jetpack 5 support for this board. Feedback on this PR is very much appreciated.

meta-mender-tegra/meta-mender-tegra-common/recipes-bsp/tegra-binaries/tegra-storage-layout_%.bbappend
```diff
@@ -1,7 +1,8 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
-SRC_URI:append:p3768-0000-p3767-0000 = " \
```
TheYoctoJester, line 3: Hi @RoopanJK,

This is the only line which I'm a bit unsure about. As the `flash_l4t_t234_nvme_rootfs_ab.xml` is now unconditionally added, does it possibly overwrite a file of the same name for other boards too? Otherwise should be good.

RoopanJK: Hi @TheYoctoJester,

I’m not entirely sure I understand your concern. Yes, the `flash_l4t_t234_nvme_rootfs_ab.xml` is now unconditionally added to the `SRC_URI` of the `tegra-storage-layout` recipe using this append file. As a result, all boards using this recipe will include this file in their resource index.

However, this won’t impact any board unless we set the `PARTITION_FILE_EXTERNAL` variable. The following lines are used to conditionally set this variable for boards that have an NVMe drive.

RoopanJK: @TheYoctoJester. Yes, the file `flash_l4t_t234_nvme_rootfs_ab.xml` will overwrite a file of the same name for other boards. But I don't see any identical files or such a situation arising.

But if it is concerning we can conditionally add it or is there a better way?. Please let me know.

Regards,
RoopanJK

RoopanJK:
Hi, I've messed up the commit messages and would like to close this dirty PR and reopen a new one. Please let me know if the proposed changes are okay. I shall create a new clean PR for JP6. 

Apologies as this is my first time trying to contribute.

TheYoctoJester:
Hi @RoopanJK,

No problem at all, contributions are always welcome! I just don't have much expertise on the Nvidia stuff so I tend to be a bit more careful. As far as I can tell right now the PR looks good with the minor requirement that all commits need a Signed-off-by of yours. Otherwise, feel free to open a clean PR for merging.
Additional thoughts maybe @pberginkonsult @dwalkes?

RoopanJK:
Opened a new PR #481.

PR #478, opened 2025-09-05
meta-mender-tegra: settings for firmware updates

pberginkonsult:
On the tegra platform firmware rollback's are blocked by bootloader edk2-nvidia. Trying to go back in version will end up in a non working update that is hard to detect the root cause of. This commit will only update firmware if the capsule shipped in the image is newer than the one installed in the next slot.

To be able to detect the installed firmware version the UEFI variable SystemFwVersions-781e084c-a330-417c-b678-38e696380cb9 is used. This variable is not documented in edk2-nvidia but what is found by test is that is reports version of firmware in slot A and version in slot B.

Version of the capsule tegra-bl.cap is put in a file tegra-bl.cap.version in build time and will be set to L4T_VERSION.

When same version is detected this will avoid to reflash same version and speed up update process.

Using this feature will make it possible to end up with non-official combination of firmware and OS. For example when doing image updates 36.4.3(A)->36.4.4(B)->36.4.4(A)->36.4.3(B) you will end up running firmware 36.4.4 with the OS from 36.4.3 as this rollback protection will take effect in the third upgrade. If this in unwanted it should be protected with provides/depends in the upgrade bundle.


pberginkonsult:
This has been tested on a Nvidia AGX Orin but no other targets. Will need more tests out there and I'm happy to get feedback if this is good way to go. The problem I'm trying to solve is to get rid of updates silently failing because images have different versions of firmware.

meta-mender-tegra/meta-mender-tegra-common/recipes-mender/tegra-state-scripts/files/switch-rootfs
```diff
@@ -49,8 +102,40 @@ tmp_mount="$(mktemp -d)"
...
+else
+	fw_version_compare=$(compare_version "$FW_VER_NEXT" "$FW_VER_1")
+fi
+
+if [ "$fw_version_compare" = "newer" ]; then
+	echo "Firmware will be updated to $FW_VER_NEXT" >&2
+	cp "${tmp_mount}"/opt/nvidia/UpdateCapsule/tegra-bl.cap /boot/efi/EFI/UpdateCapsule/TEGRA_BL.Cap
+	oe4t-set-uefi-OSIndications
+else
+	echo "Firmware will NOT be updated. Newer or same version already installed." >&2
```
dwalkes: Regarding:
> When same version is detected this will avoid to reflash same version and speed up update process.

Wouldn't the slot switch also happen if an older version is being installed and, in that case, won't this potentially cause a failure due to mismatch between bootloader and rootfs L4T versions?  Might be better to fail in this case by default, or at least leave a warning message in the logs to say "I hope you know what you are doing" or similar.

dwalkes:
Interesting idea to use `SystemFwVersions`.  You might post on the NVIDIA developer forum to ask why this is undocumented and whether you could count on it sticking around.

This looks similar to the implementation in referenced in https://github.com/orgs/OE4T/discussions/1922#discussioncomment-13522026 but with the following differences:

1. Using the undocumented UEFI variable `SystemFwVersions` to get the L4T version rather than relying on rootfs/bootloader slot alignment.
2. Restricting versioning for the update to exactly `L4T_VERSION` rather than allowing customization and ways to force capsule update within the same L4T version when you know you have a change which requires it (see related discussion [here](https://matrix.to/#/!YBfWVpJwNVtkmqVCPS:gitter.im/$BNj5OIzOmVGohYbUnCcnmKVbpIE1B5nXNbhHucxoMeQ?via=gitter.im&via=matrix.org&via=3dvisionlabs.com) in element)
3. Does slot switches by default when versions mismatch instead of only slot switching when versions exactly match and capsule updating (or at least attempting capsule update) in all other cases.

meta-mender-tegra/meta-mender-tegra-common/recipes-mender/tegra-state-scripts/files/switch-rootfs
pberginkonsult: Yes, the slot switch will happen if you install an image from older release. I tried to highlight this with this:
> Using this feature will make it possible to end up with non-official combination of firmware and OS. For example when > doing image updates 36.4.3(A)->36.4.4(B)->36.4.4(A)->36.4.3(B) you will end up running firmware 36.4.4 with the OS > from 36.4.3 as this rollback protection will take effect in the third upgrade. If this in unwanted it should be protected > with provides/depends in the upgrade bundle.

The hard requirement requirement in edk2-nvidia to never rollback version will also apply to OS image if this is not allowed. Meaning you can never OTA from 36.4.4 to 36.4.3. That is what happened to me during development. This is probably fine for product releases but when updating layers and working with development it is not optimal. But probably it is a safer way to have some kind of configuration to allow ending up in non-official combination and by default just fail here early?

pberginkonsult:
> You might post on the NVIDIA developer forum to ask why this is undocumented and whether you could count on it sticking around.

https://forums.developer.nvidia.com/t/edk2-nvidia-undocumented-variable-systemfwversions/344143 
Let's see if we can get some more info.

> 2. Restricting versioning for the update to exactly L4T_VERSION rather than allowing customization and ways to force capsule update within the same L4T version when you know you have a change which requires it (see related discussion [here](https://matrix.to/#/!YBfWVpJwNVtkmqVCPS:gitter.im/$BNj5OIzOmVGohYbUnCcnmKVbpIE1B5nXNbhHucxoMeQ?via=gitter.im&via=matrix.org&via=3dvisionlabs.com) in element)

Yes, the version in tegra-bl.cap.version, currently always L4T_VERSION, could be more flexible in its implementation to be able to override and force a capsule update, if needed. This should be reworked in this PR.

A fundamental question for me in this is still why the firmware and OS image should be that tight coupled?  I would rather see that OS image is one OTA artifact that can be updated and firmware is a separate one that can be updated through an UEFI capsule update. Are there any documentation or recommendations from Nvidia about compatibility between firmware and OS? (I guess they always will try to refuse to comment or help when you go outside of their Jetpack releases, as we anyway do with meta-tegra, but besides that.)

dwalkes:
> A fundamental question for me in this is still why the firmware and OS image should be that tight coupled?

I think it's gotten much better over the years but there are still likely at least some binary components in the OS which have dependencies to boot components that are probably somewhere between difficult and impossible to understand without insider knowledge.  Also the testing that is done by NVIDIA and pretty much everyone else will have bootloader and rootfs components aligned beween L4T releases. So if you don't align your os and boot slots with the same L4T version you are likely to be the one to find any issues which do exist and then have trouble getting someone else to care about them. If you want to decouple application content from boot/base rootfs content my guess is NVIDIA's recommendation will be to use containers given the amount of effort they've invested in this solution.  This is also possible with meta-tegra but brings its own set of challenges.

meta-mender-tegra/meta-mender-tegra-common/recipes-mender/tegra-state-scripts/files/switch-rootfs
dwalkes: > But probably it is a safer way to have some kind of configuration to allow ending up in non-official combination and by default just fail here early

Yes I'd suggest this, or at least complain loudly in the logs that this may be a problem.

TheYoctoJester:
@pberginkonsult thanks a lot for the submission and @dwalkes for the discussion. Unfortunately I don't have the expertise to judge, so what should we take as conclusion here? Merge, change, no merge?

pberginkonsult:
> @pberginkonsult thanks a lot for the submission and @dwalkes for the discussion. Unfortunately I don't have the expertise to judge, so what should we take as conclusion here? Merge, change, no merge?

I need to rewrite this and keep the current design that bootloader and rootfs versions are in sync by default. This will though block the upgrades in an earlier stage if you try to apply an upgrade which has an older bootloader. And I plan to have a configuration to allow breaking the version sync between bootloader and rootfs that can be used during development and if you know what you are doing. So don't merge this as it is now. I'll try to find some time to rework this. If you want to close it until ready I can reopen or create a new one.

TheYoctoJester:
> So don't merge this as it is now. I'll try to find some time to rework this. If you want to close it until ready I can reopen or create a new one.

Ok, thanks for the summary. I leave it to you if you want to keep this one open as mental notice.👍 Just close or adjust as you see fit.

pberginkonsult:
Reworked this PR after review comments.
Default behavior kept if no config changes are actively done.
Two new variables `TEGRA_MENDER_INSTALL_ONLY_IF_DIFFERENT` and `TEGRA_MENDER_ALLOW_FIRMWARE_VERSION_MISSMATCH` that can be used to tune the behavior. Tried to document the usage of them in `tegra-mender-setup.bbclass`.
Please review and comment.

dwalkes:
I haven't fully parsed the changes here but one note is that the approach to use `SystemFwVersions` will only work on Jetpack 6 I believe, unless you backport https://github.com/NVIDIA/edk2-nvidia/commit/963acbf0fdf4ac06859429d4ade9ec31069f072d.  I don't see this var on my Jetpack 5 build.

PR #481, opened 2025-09-18
scarthgap: add jetson-orin-nano-devkit-nvme board - Jetpack 6

RoopanJK:
1. Add Jetpack 6 support for board `jetson-orin-nano-devkit-nvme`.
2. Fix typo in partition 17's name in layout file `flash_l4t_t234_nvme_rootfs_ab.xml`.

### DIFF
--- a/kas/tegra/jetpack6/jetson-orin-nano-devkit-nvme.yml
+++ b/kas/tegra/jetpack6/jetson-orin-nano-devkit-nvme.yml
@@ -1,0 +1,21 @@
+header:
+  version: 14
+  includes:
+    - kas/include/mender-full.yml
+    - kas/include/tegra-jetpack6.yml
+
+machine: jetson-orin-nano-devkit-nvme
+
+local_conf_header:
+  AB-upgrades: |
+    UBOOT_EXTLINUX = "1"
+    USE_REDUNDANT_FLASH_LAYOUT_DEFAULT = "1"
+  eMMC: |
+    TEGRAFLASH_NO_INTERNAL_STORAGE = "1"
+    # Mender tegra_mender_calc_total_size workaround
+    # Setting MENDER_STORAGE_TOTAL_SIZE_MB to 16Gb and no EMMC
+    EMMC_SIZE = "0"
+    MENDER_STORAGE_TOTAL_SIZE_MB = "16384"
+  Permanent Data Partition: |
+    MENDER_DATA_PART_NUMBER:tegra = "17"
+    MENDER_STORAGE_DEVICE:tegra = "/dev/nvme0n1"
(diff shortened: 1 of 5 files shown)


kas/tegra/jetpack6/jetson-orin-nano-devkit-nvme.yml
```diff
@@ -0,0 +1,21 @@
...
+    - kas/include/tegra-jetpack6.yml
+
+machine: jetson-orin-nano-devkit-nvme
+
+local_conf_header:
+  AB-upgrades: |
+    UBOOT_EXTLINUX = "1"
+    USE_REDUNDANT_FLASH_LAYOUT_DEFAULT = "1"
+  eMMC: |
+    TEGRAFLASH_NO_INTERNAL_STORAGE = "1"
```
dwalkes: Is this needed?  I'd expect you could defer to the MACHINE settings here.  I suspect because the devkit has a SDCard slot it's not setting `TEGRAFLASH_NO_INTERNAL_STORAGE` in the MACHINE but not sure about this.  What happens if this isn't specified?

dwalkes (commented):
I haven't tried myself but one comment here based on a quick review.

kas/tegra/jetpack6/jetson-orin-nano-devkit-nvme.yml
RoopanJK: After you pointed out, I did track it and found [this](https://github.com/OE4T/meta-tegra/blob/aca8ae6dcaedd1651337838514d0797059c41406/recipes-bsp/tegra-binaries/tegra-storage-layout_36.4.4.bb#L51-L60) which is self-explanatory. But then I did test by not setting the `TEGRAFLASH_NO_INTERNAL_STORAGE` and the flash succeeded on the devkit with NVMe.

I actually set the variable with reference to other boards that set `PARTITION_LAYOUT_EXTERNAL_DEFAULT` such as [p3768-0000-p3767-0003](https://github.com/OE4T/meta-tegra/blob/aca8ae6dcaedd1651337838514d0797059c41406/conf/machine/p3768-0000-p3767-0003.conf#L15-L16) and [p3768-0000-p3767-0004](https://github.com/OE4T/meta-tegra/blob/3bf6f0705b141c671ade9f155505a48682384a2e/conf/machine/p3768-0000-p3767-0004.conf#L10-L11) to be safe, but both of these boards do not have an SD card slot.

But I'm still kinda puzzled on how the `TEGRAFLASH_NO_INTERNAL_STORAGE` affect the board.

dwalkes: I think today it's only the logic which allows use of an internal flash config for the sdcard, which you aren't using here anyway, so I don't think it really matters either way.  My hesitation to include it would just be in case it does matter for some reason later, would rather defer to MACHINE settings unless there's some specific reason you need to use something different.

RoopanJK: Yeah thought so. Okay I'll remove that variable and defer to the default MACHINE settings. Thank you @dwalkes

PR #485, opened 2025-10-23
Remove artifact collection from GitHub Actions workflows

TheYoctoJester:
This removes the artifact collection steps from build_boards, build_demos, and build_validation workflows to simplify the CI pipeline.

The functionality did not work out sufficiently and will be replaced with a different approach soon, this cleans up in preparation.

### DIFF
--- a/.github/workflows/build_boards.yml
+++ b/.github/workflows/build_boards.yml
@@ -12,15 +12,10 @@ env:
 jobs:
   prepare:
     runs-on: [self-hosted, linux, x64, builder]
-    outputs:
-      artifact_path: ${{ steps.artifact_path.outputs.artifact_path }}
     steps:
       - name: Clean up
         run: rm -fR $BUILDDIR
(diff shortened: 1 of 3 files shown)


oldgiova (approved):
>  this cleans up in preparation.

Approving because I trust your judgement here, but since this was needed, I wonder if this the correct approach; I mean without the fix ready, isn't this cleanup creating a non-working process?

TheYoctoJester:
> > this cleans up in preparation.
> 
> Approving because I trust your judgement here, but since this was needed, I wonder if this the correct approach; I mean without the fix ready, isn't this cleanup creating a non-working process?

We've never officially provided any artifacts from these builds, and this was my attempt at creating a way to do it. However it never worked correctly, so the GitHub Actions approach is for now going back to only build testing, and I'm currently building a fully different thing on a self-hosted ForgeJo instance.

So no worries, this is not taking away anything @oldgiova 👍

PR #491, opened 2026-01-21
Fix: patch append for more target certainty

Levitiku5:
Sponsored by 21SoftWare

My team and I make use of the `meta-mender-raspberrypi` layer, but we noticed that RPi-specific patches were still being searched for when the target was not an RPi:

```
meta-lts-mixins/recipes-bsp/u-boot/u-boot_2025.04.bb: Unable to get checksum for u-boot SRC_URI entry 0001-fix-disable-bootdelay-keypress-detection-on-Pi5-fami.patch: file could not be found
```

This commit appends the patch while guarded by machine-specific syntax, in line with how other patches are appended. This adds certainty and prevents the error detailed above.


21sw-clayton:
> Commit 0d525f53e65890d435ae2bd1e89754c9b62015da is not signed off! Use --signoff with your commit. Make sure that the Author of the commit matches the one in Signed-off-by

TheYoctoJester (approved):

PR #504, opened 2026-04-12
feat: Add qemux86-64 U-Boot EFI demo with Mender A/B support

TheYoctoJester:
Add a QEMU-testable demo that replaces GRUB with U-Boot running as an EFI application under OVMF on x86-64, providing a GPLv3-free boot chain with Mender A/B rootfs updates.

Boot chain: QEMU -> OVMF -> u-boot-app.efi (from ESP) -> Linux kernel

The implementation uses a 5-partition GPT layout with a dedicated FAT partition for the U-Boot environment, separate from the ESP, to prevent FAT corruption that breaks OVMF boot discovery after guest reboots.

Notable adaptations for EFI app mode:
- SATA/AHCI storage instead of virtio (OVMF reboot re-enumeration)
- Inline altbootcmd recovery instead of reset (OVMF exits to UI)
- Conditional kernel load (load && zboot) to avoid stale RAM boots
- Pre-populated env via mkenvimage with redundant env support

Includes a kas overlay for the bootloader validation suite.

Based on: https://hub.mender.io/t/u-boot-efi-on-intel-corei7-64/8216

### DIFF
--- a/kas/demos/meta-mender-x86-uboot-efi/LICENSE
+++ b/kas/demos/meta-mender-x86-uboot-efi/LICENSE
@@ -1,0 +1,202 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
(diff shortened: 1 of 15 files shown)


TheYoctoJester:
Superseded by the wrynose port. The qemux86-64 U-Boot EFI demo now lives as the `meta-mender-x86-uboot-efi` layer (plus its `meta-mender-explicit-wic` dependency) on the `wrynose-demos` branch of theyoctojester/meta-mender-community, with the `qemux86-64-uboot-efi.yml` build config on the `wrynose-demos` branch of theyoctojester/mender-community-images. The u-boot patch was rebased onto 2026.01 and the full A/B OTA loop verified end-to-end against hosted.mender.io. Closing this scarthgap PR in favour of that.

PR #509, opened 2026-04-22
fix: relocate firmware DTB to fdt_addr_r before booti on RPi

lhoward:
U-Boot 2026.01 tightened lmb_reserve() in image_setup_libfdt(). The firmware-placed DTB (parked by start4.elf near the top of ARM-visible RAM) now fails reservation:

    Working FDT set to 3af36f30
    Failed to reserve memory for fdt at 0x3af36f30
    FDT creation failed!
    resetting ...

Move the blob to fdt_addr_r (a fixed low slot defined in rpi_arm64's ENV_MEM_LAYOUT_SETTINGS) and hand that address to booti instead. This is the same pattern meta-mender's generic boot code already uses for non-RPi targets, so it aligns RPi with the rest of the ecosystem rather than introducing anything new. Works on all CM4 variants (1/2/4/8 GB); the symptom actually grows worse on larger parts because start4.elf places the DTB proportionally higher.

Verified on a 1 GB CM4 with U-Boot 2026.01 and linux 6.18.

AI disclosure: Claude Code.

### DIFF
--- a/meta-mender-raspberrypi/recipes-bsp/rpi-u-boot-scr/files/boot.cmd.in
+++ b/meta-mender-raspberrypi/recipes-bsp/rpi-u-boot-scr/files/boot.cmd.in
@@ -2,5 +2,11 @@ fdt addr ${fdt_addr} && fdt get value bootargs /chosen bootargs
 run mender_setup
 mmc dev ${mender_uboot_dev}
 load ${mender_uboot_root} ${kernel_addr_r} /boot/@@KERNEL_IMAGETYPE@@
-@@KERNEL_BOOTCMD@@ ${kernel_addr_r} - ${fdt_addr}
+# Relocate the firmware-placed DTB to fdt_addr_r before booti. U-Boot
+# >= 2026.01 tightened lmb_reserve() in image_setup_libfdt() and rejects
+# the firmware's high placement (start4.elf parks the DTB near the top
+# of ARM-visible RAM). fdt_addr_r is a fixed low slot defined in
+# rpi_arm64's ENV_MEM_LAYOUT_SETTINGS and is safe on all CM4 variants.
+fdt move ${fdt_addr} ${fdt_addr_r}
+@@KERNEL_BOOTCMD@@ ${kernel_addr_r} - ${fdt_addr_r}
 run mender_try_to_recover


TheYoctoJester:
Hi @lhoward,

Thanks for the submission! I do test on RPi4 and RPi5, those seem to be not affected, so I wonder a bit why the CM4 is. Thoughts maybe @leon-anavi?

lhoward:
Yeah, a bit about my pay grade. I'll check we don't have any overrides that might be causing this. Wrynose branches (of everything) with older U-Boot were fine.

leon-anavi:
> Hi @lhoward,
> 
> Thanks for the submission! I do test on RPi4 and RPi5, those seem to be not affected, so I wonder a bit why the CM4 is. Thoughts maybe @leon-anavi?

Sorry, given the similarities between the Raspberry Pi 4 and the Compute Module 4 in this regard, nothing immediately comes to mind.

lhoward:
I'll do some more research and update.

lhoward:
Report back from @claude.

Root cause is U-Boot **v2025.10**'s LMB rework (Sughosh Ganu / Linaro, released 2025-10-06), which started honoring the return value of `lmb_alloc_mem(LMB_MEM_ALLOC_ADDR, …)` in `boot_relocate_fdt()`. Prior releases called the old `lmb_reserve()` but **ignored the error**, so the same condition was silently tolerated.

In Yocto terms, this first lands in:

- OE-core commit [`237d980c11`](https://git.openembedded.org/openembedded-core/commit/?id=237d980c11) — *"u-boot: upgrade 2025.07 -> 2025.10"*
- First tagged release: **`yocto-5.3` (Whinlatter)**
- Persists into **`yocto-6.0` (wrynose)** with U-Boot 2026.01

The failing code path (`boot/image-fdt.c` in 2026.01):

```c
err = lmb_alloc_mem(LMB_MEM_ALLOC_ADDR, 0, &addr, of_len, LMB_NONE);
if (err) {
    printf("Failed to reserve memory for fdt at %#llx\n", (u64)addr);
    goto error;
}
```

Any RPi with **>512 MB of ARM-visible RAM** should be affected, because `start4.elf` parks the firmware DTB near the top of ARM-visible RAM — well above `bootm_size=0x20000000` (512 MB) defined in `rpi.env`. Nothing CM4-specific is going on here; the same condition applies to any Pi4/CM4 variant.

If your RPi4 / RPi5 CI is on **scarthgap** (U-Boot 2024.01) or **walnascar** (2025.01), the regression is invisible — both are pre-rework. Would you be able to retest against **wrynose + U-Boot 2026.01** specifically? That's the only combination where this surfaces.

### References
- [U-Boot v2025.10 release announcement (LWN)](https://lwn.net/Articles/1041023/) — *"The LMB rework … is now here"*
- [u-boot ML thread: *"riscv: v2025.10-rc2 fails on QEMU with 'Failed to reserve memory for fdt'"*](https://www.mail-archive.com/u-boot@lists.denx.de/msg552344.html) — identical failure mode on a different arch
- [Ganu's LMB rework patch series (v2)](https://lore.kernel.org/u-boot/20240814110009.45310-5-sughosh.ganu@linaro.org/T/)

danielskinstad (approved):
I saw the same errors when attempting to update from scarthgap to wrynose. I've confirmed that it works locally

PR #516, opened 2026-06-23
Add `wrynose` meta-tegra support, Jetpack 7 layer

Qapf:
Imports `meta-mender-tegra` from scarthgap at https://github.com/mendersoftware/meta-mender-community/commit/bc815f010f3b6c2180c2acb2b632bdcc9ff992c6

Swaps all `${S}` uses for `${UNPACKDIR}`

Add `meta-mender-tegra-jetpack7`, a JetPack 7 layer targeting the WIP 39.2.0 branch. 

Tested on a thor devkit.


TheYoctoJester:
I ran a local verification build of this PR for **JetPack 7 / Jetson AGX Thor** and it builds end to end — nice work getting Thor support together. A few notes from the exercise, mostly about reproducibility for others who want to build the JP7 target.

## Build result

`core-image-minimal` for `MACHINE = jetson-agx-thor-devkit`, all 6776 tasks succeeded, producing the expected `.tegraflash-tar`, `.mender`, and `.dataimg` artifacts. (Built only — not yet flashed/boot-tested on a Thor devkit.)

Layer revisions used:

- `meta-tegra` / `meta-tegra-community` / `tegra-demo-distro` @ `wip-l4t-r39.2.0`
- `openembedded-core` / `bitbake` / `meta-openembedded` / `meta-virtualization` @ the `master` revisions `tegra-demo-distro` `wip-l4t-r39.2.0` pins
- `meta-mender` @ `wrynose`

## 1. `LAYERSERIES_COMPAT` vs. the actual base for JP7

The JP7 layers declare `LAYERSERIES_COMPAT = "wrynose"`, but the JP7 / Thor stack can't actually be built on a wrynose base:

- `jetson-agx-thor-devkit` only exists on meta-tegra's `wip-l4t-r39.2.0` branch (the `wrynose` branch of meta-tegra has no Thor machine).
- meta-tegra/`wip-l4t-r39.2.0`, meta-tegra-community/`wip-l4t-r39.2.0`, and tegra-demo-distro/`wip-l4t-r39.2.0` all declare `LAYERSERIES_COMPAT = "blacksail"` and are built against oe-core/bitbake/meta-oe/meta-virt on `master`.

The build only succeeds because the oe-core `master` revision the demo-distro pins has `LAYERSERIES_CORENAMES = "wrynose blacksail"`, so the `wrynose`-tagged Mender layers and the `blacksail`-tagged tegra layers happen to both pass bitbake's compat check. That's a fairly narrow window. It may be worth either bumping the JP7 layer's `LAYERSERIES_COMPAT` to include `blacksail`, or being explicit in the README that JP7 targets the blacksail/master era rather than wrynose.

## 2. No `kas` configuration for JetPack 7

The README's kas section lists configs for JP5 and JP6, but there's no `kas/tegra/jetpack7/` board file (and no Thor entry anywhere), so there's no turnkey way to build the new target. Adding a `kas/tegra/jetpack7/jetson-agx-thor-devkit.yml` (plus a `tegra-jetpack7.yml` include pinning the `wip-l4t-r39.2.0` revisions) in line with the existing JP5/JP6 layout would make the JP7 path reproducible.

## 3. Container-build note (upstream meta-tegra, not this PR)

When building under `kas-container` (non-root build uid), `meta-tegra`'s `tegra-storage-layout-base` installs the prebuilt L4T flash-layout XML verbatim, so the file ends up owned by the build uid. Under pseudo this trips `OEOuthashBasic` ("`getpwuid(): uid not found`") and then the fatal `host-user-contaminated` QA. This is an upstream meta-tegra × containerized-build interaction rather than anything in this PR — flagging it only because anyone running the JP7 build in a container (e.g. CI) will hit it.

Happy to share the kas config I wrote for JP7/Thor if that's useful for the PR.

------

Full disclosure: Claude Code was involved in both testing and writing this comment.

Qapf:
> I ran a local verification build of this PR for JetPack 7 / Jetson AGX Thor and it builds end to end — nice work getting Thor support together. A few notes from the exercise, mostly about reproducibility for others who want to build the JP7 target.

Thank you for your kind words.

> Build result & LAYERSERIES_COMPAT vs. the actual base for JP7

I can confirm this is the same exact set of pins I am maintaining to make this work. 

In a sentence, I look at what `tegra-demo-distro` pins in a given successful build of `wip-l4t-r39.2.0` and use those commits. 

RE: `LAYERSERIES_CORENAMES`. It may be worth waiting on final merge of this PR until the wip `meta-tegra` branch [merges into master](https://github.com/OE4T/meta-tegra/pull/2247) and the shape of `meta-tegra`'s long term support for 39.2.0 becomes clear.

> `kas` configuration

I didn't try to bite this off as the `wrynose` branch didn't have `kas` defined for any board. Latest now includes a kas config for `jetson-agx-thor-devkit`

I am able to build to success with

```
docker run -u 1000:1000 -v /path/to/meta-mender-community:/builder -it ghcr.io/siemens/kas/kas:latest
kas build ./kas/tegra/jetpack7/jetson-agx-thor-devkit.yaml
```

> host-user-contaminated

I added `chown -R root:root ${D}` to the end of the `do_install` in `tegra-storage-layout-base_%.bbappend` so that the files are owned by the pesudo-root within a container. If you want me to drop it, let me know, but I do think the average person who might use this is probably building in a container, and that should just work.

Qapf:
`meta-mender` finished backporting Jetpack 7.2 to `wrynose` yesterday. 

I've updated the kas manifests to point at `wrynose`

the manifests builds to success for me with

```
docker run -u 1000:1000 -v /path/to/meta-mender-community:/builder -it ghcr.io/siemens/kas/kas:latest
kas build ./kas/tegra/jetpack7/jetson-agx-thor-devkit.yaml
```

meta-mender-tegra/meta-mender-tegra-common/classes-global/tegra-mender-setup.bbclass
```diff
@@ -0,0 +1,178 @@
...
+            bb.error('TEGRAFLASH_SDCARD_SIZE does not end with G, K, or M')
+    else:
+        emmc_size = d.getVar('EMMC_SIZE')
+        if not emmc_size:
+            # No internal storage (e.g. t264 NVMe-only platforms): the A/B rootfs
+            # and data partition sizes come from the L4T flash layout XML, not from
+            # a computed eMMC total. Return a large nominal total so mender's
+            # rootfs-fits bookkeeping passes (the distro can still pin an explicit
+            # MENDER_STORAGE_TOTAL_SIZE_MB) instead of aborting the build.
+            return 196608
```
njkyu-hex: Is the size chosen totally arbitrary?

Qapf: yep. I picked a number out of thin air. If there's a better way to handle eMMC-less platforms I am happy to change.

TheYoctoJester:
I re-ran the local verification build at the current head (c94c318) — `core-image-minimal` for `jetson-agx-thor-devkit` now builds clean against the real `wrynose` pins, 6640/6640 tasks, with the expected `.tegraflash-tar`, `.mender` and `.dataimg` artifacts. All the points from my earlier comment are addressed, thanks for the quick turnaround.

One new observation from inspecting the resulting images, this time at the runtime level rather than the build level:

## `MENDER_STORAGE_DEVICE` on the NVMe-only Thor

`MENDER_STORAGE_DEVICE` falls back to meta-mender's default `/dev/mmcblk0`, and that propagates into the built artifacts:

- the persistent `mender.conf` on the data partition carries `"RootfsPartA": "/dev/mmcblk0p1"` and `"RootfsPartB": "/dev/mmcblk0p2"`
- fstab mounts `/data` from `/dev/mmcblk0p11`

Since the Thor devkit boots NVMe-only (`EMMC_SIZE` is empty — the very reason the eMMC-less code path exists), none of these device nodes are present at runtime: the `/data` mount fails at boot, and a dual-rootfs deployment cannot locate the inactive partition. `tegra-mender-setup.bbclass` already has a per-machine override for the Orin Nano devkit, so the analogous

```
MENDER_STORAGE_DEVICE_DEFAULT:tegra264 = "/dev/nvme0n1"
```

should cover it (meta-mender derives the `pN` partition suffixes correctly for device names ending in a digit).

Caveat: found by inspecting the images with debugfs, not on hardware — I don't have a Thor devkit at hand. If your on-device testing included a full deployment cycle with these defaults, I'd be curious how it passed.

Full disclosure: Claude Code was involved in both testing and writing this comment.

meta-mender-tegra/meta-mender-tegra-common/classes-global/tegra-mender-setup.bbclass
TheYoctoJester: One option grounded in existing metadata: meta-tegra's Thor machine includes already set `ROOTFSPART_SIZE_DEFAULT` (50GiB for `agx-thor-t5000`/`t4000`), so the no-eMMC path could derive the nominal total from that instead — roughly `2 * ROOTFSPART_SIZE_REDUNDANT + MENDER_DATA_PART_SIZE_MB + reserved`. Per-slot sizes would then track whatever upstream sets per machine. For reference, with the current constant the generated layout ends up with two ~96GiB APP partitions (103012106240 bytes each in `external-flash.xml.in`) and a ~92GiB sparse rootfs image per build.

That said, keeping the constant as-is would also be fine in my view: it is only a `_DEFAULT`, a distro can pin `MENDER_STORAGE_TOTAL_SIZE_MB` explicitly, and the devkit's 1TB NVMe has plenty of headroom.

Qapf:
> I re-ran the local verification build at the current head ([c94c318](https://github.com/mendersoftware/meta-mender-community/commit/c94c318179ff461f571d5d62d8a99df38706b8bc)) — `core-image-minimal` for `jetson-agx-thor-devkit` now builds clean against the real `wrynose` pins, 6640/6640 tasks, with the expected `.tegraflash-tar`, `.mender` and `.dataimg` artifacts. All the points from my earlier comment are addressed, thanks for the quick turnaround.
> 
> One new observation from inspecting the resulting images, this time at the runtime level rather than the build level:
> ## `MENDER_STORAGE_DEVICE` on the NVMe-only Thor
> 
> `MENDER_STORAGE_DEVICE` falls back to meta-mender's default `/dev/mmcblk0`, and that propagates into the built artifacts:
> 
>     * the persistent `mender.conf` on the data partition carries `"RootfsPartA": "/dev/mmcblk0p1"` and `"RootfsPartB": "/dev/mmcblk0p2"`
> 
>     * fstab mounts `/data` from `/dev/mmcblk0p11`
> 
> 
> Since the Thor devkit boots NVMe-only (`EMMC_SIZE` is empty — the very reason the eMMC-less code path exists), none of these device nodes are present at runtime: the `/data` mount fails at boot, and a dual-rootfs deployment cannot locate the inactive partition. `tegra-mender-setup.bbclass` already has a per-machine override for the Orin Nano devkit, so the analogous
> 
> ```
> MENDER_STORAGE_DEVICE_DEFAULT:tegra264 = "/dev/nvme0n1"
> ```
> 
> should cover it (meta-mender derives the `pN` partition suffixes correctly for device names ending in a digit).
> 
> Caveat: found by inspecting the images with debugfs, not on hardware — I don't have a Thor devkit at hand. If your on-device testing included a full deployment cycle with these defaults, I'd be curious how it passed.
> 
> Full disclosure: Claude Code was involved in both testing and writing this comment.

Fixed in latest. `MENDER_STORAGE_DEVICE_DEFAULT:tegra264` is now `/dev/nvme0n1`, as you
suggested. Bedrock's distro sets the storage device in its own layer, which masked
this: the on-device testing I reported ran with those variables in place, so the
community defaults were never exercised on hardware and would have failed the way you
describe.

`mender.conf` on the data partition now carries `/dev/nvme0n1p1` and `/dev/nvme0n1p2`,
fstab mounts `/data` from `/dev/nvme0n1p11`, and no `mmcblk` references remain in the
rootfs or the generated flash layout.

meta-mender-tegra/meta-mender-tegra-common/classes-global/tegra-mender-setup.bbclass
Qapf: Fixed in latest. The partition is now sized from `ROOTFSPART_SIZE_REDUNDANT` as you suggested. 

The no-eMMC path now returns two of those slots plus the data, boot and swap partitions, and `196608`
stays only as a fallback for a machine that sets no `ROOTFSPART_SIZE_DEFAULT`. 

The guard is on the latter because `ROOTFSPART_SIZE_REDUNDANT` is
`int(ROOTFSPART_SIZE_DEFAULT) // 2` in `tegra-common.inc`, so it cannot expand at all
when `_DEFAULT` is unset.

I left out any reserved-space subtraction. On this path the NVIDIA partitions come from
the flash layout XML and do not compete with mender's total, unlike the eMMC path which
still subtracts `TEGRA_MENDER_RESERVED_SPACE_MB`.

On `jetson-agx-thor-devkit` `MENDER_STORAGE_TOTAL_SIZE_MB` drops from 196608 to 51328,
giving 25GiB `APP` and `APP_b` partitions in place of the 96GiB ones you measured. That
is NVIDIA's stock slot size, since `ROOTFSPART_SIZE_DEFAULT` is 50GiB for
`agx-thor-t5000`/`t4000`, so per-slot sizes now track whatever upstream sets per machine.

TheYoctoJester:
Thanks @Qapf!

PR #517, opened 2026-07-07
meta-mender-qcom: Arduino Uno Q Mender integration (A/B rootfs + STM32U585 MCU firmware)

TheYoctoJester:
## Summary

Adds `meta-mender-qcom`, a new Mender integration layer for Qualcomm
(`qualcomm-linux/meta-qcom`) based boards, with the **Arduino Uno Q**
(Dragonwing QRB2210, MACHINE `uno-q`) as the first supported and hardware-tested
board.

Rather than Mender's classic dual-rootfs + U-Boot/GRUB integration, this layer
reuses the platform's **native Qualcomm A/B boot slots** (ABL + `qbootctl`) and
drives them from a custom Mender **Update Module** (`qbootctl-rootfs`). This fits
the Qualcomm boot flow (PBL → XBL → ABL → U-Boot → UEFI → systemd-boot/UKI)
without re-partitioning around Mender's model.

## What the layer provides

- **Slotted OS layout** — a `qcom-partition-conf` bbappend derives a
  `system_a`/`system_b` (+ `userdata`, + `dtbo_a`/`dtbo_b`, which `qbootctl -s`
  hard-requires) GPT from the stock `qrb2210-unoq` table.
- **Slot-aware, self-healing initramfs** (`unoq-initramfs-ab` + the `86-abslot`
  initramfs-framework module) — reads the active slot via `qbootctl`, mounts
  `system_<slot>`; if the active slot has no valid filesystem it switches to the
  other slot (`qbootctl -s`/`-m`) and reboots, so a bad update rolls back
  deterministically.
- **`qbootctl-rootfs` Update Module** — writes the payload to the inactive slot,
  activates it, reboots (Automatic), verifies the booted slot, commits with
  `qbootctl -m`, rolls back on failure.
- **Persistent Mender state** — `userdata` is mounted at `/data` and
  `/var/lib/mender` is bound onto it so device identity + in-flight update state
  survive the rootfs swap.
- **bless-boot gating** — a drop-in suppresses meta-qcom's automatic slot-bless
  while a Mender update is in flight so Mender owns commit/rollback; normal boots
  still auto-bless.

## Testing

Verified end-to-end on hardware (Arduino Uno Q) against Hosted Mender:
- A/B rootfs OTA deployment → success, device boots the updated slot,
  `artifact_name` updated.
- Deliberately broken artifact → initramfs self-heals to the known-good slot,
  Mender reports the deployment as Failure, device stays on the good slot.

## Dependencies

Depends on `meta-qcom` (qualcomm-linux, `master`) and `meta-qcom-3rdparty`
(`main`, which provides the `uno-q` MACHINE + partition tables + firmware/kernel/
u-boot), plus `meta-mender` (`wrynose`).

## Companion build config

The kas build config lives in a companion repo (separate PR):
https://github.com/theyoctojester/mender-community-images — `yocto/wrynose/floating/uno-q.yml`.
The two are coupled; that config tracks this repo's `wrynose` branch, so it
builds once this PR merges.

## Notes / open questions (why this is a draft)

- **Layer name** `meta-mender-qcom` follows the repo convention of naming after
  the upstream BSP family (Uno Q is a MACHINE under it) — happy to rename if you'd
  prefer e.g. per-board.
- **Base branch** `wrynose`. It carries no kas/CI machinery, which is why the build
  config lives in `mender-community-images` rather than here — confirm that split
  matches your intent.
- **Connectivity is intentionally not shipped** (no baked credentials); WiFi/network
  is site-specific and configured at runtime or via a local overlay.
- The Uno Q build uses meta-qcom's `nodistro` flow (oe-core master, meta-qcom
  master, meta-qcom-3rdparty `main`) rather than the poky/`core-image-minimal`
  flow other boards use — reflected in the companion kas config.
- **Future work** (designed-for, not in this PR): the same Update-Module approach
  extends to the on-board **STM32U585 MCU** firmware (Zephyr) and to **AI models**
  shipped with Edge Impulse, as additional Mender artifact types beside the rootfs one.

### DIFF
--- a/meta-mender-qcom/recipes-core/initramfs/files/86-abslot
+++ b/meta-mender-qcom/recipes-core/initramfs/files/86-abslot
@@ -1,0 +1,71 @@
+#!/bin/sh
+# initramfs-framework module: Uno Q A/B slot-aware root selection.
+# Runs after 01-udev (so /dev/disk/by-partlabel exists) and before 90-rootfs.
+# Reads the active slot from the GPT AB attribute bits via qbootctl (which falls
+# back to the GPT active bit when no androidboot.slot_suffix is present) and
+# points bootparam_root at system_<suffix>, which 90-rootfs then mounts.
+# Module name (init parses field 2 of the filename): "abslot".
+
+abslot_enabled() {
+    command -v qbootctl >/dev/null 2>&1
+}
+
+abslot_run() {
+    _out=$(qbootctl -c 2>/dev/null || true)
+    # qbootctl -c prints e.g. "Current slot: _a"
+    _suffix=$(printf '%s\n' "$_out" | sed -n 's/.*[Cc]urrent slot:[[:space:]]*//p')
+    case "$_suffix" in
+        _a|_b) ;;
+        *) _suffix=_a ;;   # safe default if the slot can't be determined
+    esac
+
+    _dev="/dev/disk/by-partlabel/system${_suffix}"
+    _i=0
+    while [ ! -e "$_dev" ] && [ "$_i" -lt 10 ]; do
+        sleep 1
+        _i=$((_i + 1))
+    done
+
+    # If the active slot holds no valid filesystem (e.g. a failed/garbage OTA to
+    # the inactive slot that was then activated), self-heal: switch the active
+    # slot to the other one via qbootctl and reboot. A single, deterministic
+    # fallback into the known-good slot. Afterwards qbootctl -c reflects the other
+    # slot, so Mender's ArtifactVerifyReboot sees we're not on the update's target
+    # slot and rolls the deployment back. Guarded with retries so a transient
+    # (device not ready) never triggers it.
+    # Authoritative validity test: actually try to mount the slot read-only.
+    # (blkid is unreliable here: a failed OTA may zero only the start of the
+    # partition, leaving backup superblocks that blkid still recognises while the
+    # real mount fails.)
+    _ok=""
+    _j=0
+    mkdir -p /run/abprobe
+    while [ "$_j" -lt 5 ]; do
+        if mount -o ro "$_dev" /run/abprobe >/dev/null 2>&1; then
+            umount /run/abprobe >/dev/null 2>&1 || true
+            _ok=1
+            break
+        fi
+        sleep 1
+        _j=$((_j + 1))
+    done
+    if [ -z "$_ok" ]; then
+        _other=1
+        [ "$_suffix" = "_a" ] || _other=0
+        printf 'unoq-ab: slot %s unmountable; switching to slot %s and rebooting\n' "$_suffix" "$_other" > /dev/kmsg 2>/dev/null || true
+        # Activate the known-good slot AND mark it successful, so ABL keeps it
+        # (qbootctl -s clears the successful bit; without -m, and with bless-boot
+        # gated by the update flag, ABL could roll the fallback back into the bad
+        # slot -> loop). This is the boot-level rollback commit.
+        qbootctl -s "$_other" >/dev/null 2>&1 || true
+        qbootctl -m "$_other" >/dev/null 2>&1 || true
+        sync
+        sleep 1
+        reboot -f >/dev/null 2>&1 || true
+        echo b > /proc/sysrq-trigger 2>/dev/null || true
+        sleep 30
+    fi
+
+    bootparam_root="PARTLABEL=system${_suffix}"
+    printf 'unoq-ab: active slot %s -> root=%s\n' "$_suffix" "$bootparam_root" > /dev/kmsg 2>/dev/null || true
+}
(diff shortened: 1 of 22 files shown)


TheYoctoJester:
Extended the layer with a **second Mender update type**: `zephyr-mcu`, which flashes the Uno Q's on-board STM32U585 MCU firmware over the QRB2210's internal SWD link (OpenOCD + `linuxgpiod`) — SWD read-back verify, `NeedsArtifactReboot=No`, backup/re-flash rollback. Adds the OpenOCD `linuxgpiod` bbappend, the `arduino-uno-q` Cortex-M33 MACHINE + `mcu` multiconfig, and a `zephyr-unoq-blink` demo firmware under `dynamic-layers/zephyrcore/` (so the layer keeps no hard meta-zephyr dependency). Verified on hardware via Hosted Mender: MCU OTA success + rollback. Companion build config: theyoctojester/mender-community-images#2.

meta-mender-qcom/recipes-core/initramfs/files/86-abslot
```diff
@@ -0,0 +1,71 @@
+#!/bin/sh
+# initramfs-framework module: Uno Q A/B slot-aware root selection.
+# Runs after 01-udev (so /dev/disk/by-partlabel exists) and before 90-rootfs.
+# Reads the active slot from the GPT AB attribute bits via qbootctl (which falls
+# back to the GPT active bit when no androidboot.slot_suffix is present) and
```
vpodzime: `androidboot.slot_suffix`? I don't understand how that's related. This part of the comment is quite confusing for me.

meta-mender-qcom/recipes-support/unoq-ab-integration/files/qbootctl-rootfs
```diff
@@ -0,0 +1,92 @@
...
+other_num() { [ "$1" = "0" ] && echo 1 || echo 0; }
+slot_dev() { [ "$1" = "1" ] && echo /dev/disk/by-partlabel/system_b || echo /dev/disk/by-partlabel/system_a; }
+
+case "$STATE" in
+    Download)
+        # Stream the payload straight onto the inactive system slot.
+        tgt=$(other_num "$(cur_num)")
+        dev=$(slot_dev "$tgt")
+        f="$(cat stream-next)"
+        [ -n "$f" ] && cat "$f" > "$dev"
```
vpodzime: I'm not sure `cat` is the best tool for this. Isn't `dd` generally available too? And what about `mender-flash`? That could save some flash writes, at least.

meta-mender-qcom/recipes-support/unoq-ab-integration/files/unoq-mender-persist
```diff
@@ -0,0 +1,20 @@
...
+set -e
+
+SRC=/var/lib/mender     # baked rootfs dir (before bind)
+DST=/data/mender        # persistent store on the userdata partition
+
+mkdir -p "$DST"
+
+# Always refresh the config from the image (server URL / tenant token), but
+# preserve identity + state that live only on /data.
+[ -e "$SRC/mender.conf" ] && cp -a "$SRC/mender.conf" "$DST/mender.conf"
```
vpodzime: I don't think we want to do this on every (re)boot, do we?

vpodzime (commented):
Looks good to me otherwise.

meta-mender-qcom/recipes-core/initramfs/files/86-abslot
TheYoctoJester: Fair point, that was confusingly worded. Reworded in a06630f: qbootctl prefers an Android-style `androidboot.slot_suffix=` kernel argument, but this board's U-Boot flow passes none, so it derives the slot from the A/B attribute bits in the GPT partition entries — i.e. the slot ABL selected. That fallback is the path in effect here (visible at runtime as qbootctl's "Couldn't find cmdline arg: 'slot_suffix'" notice in the deployment logs).

meta-mender-qcom/recipes-support/unoq-ab-integration/files/qbootctl-rootfs
TheYoctoJester: Good call — switched to `mender-flash` in a06630f, mirroring the stock `rootfs-image` module: the module now answers Yes to `ProvidePayloadFileSizes` and handles `DownloadWithFileSizes` (mender-flash needs the exact input size for a streamed payload), keeping a `cat` fallback, and `mender-flash` is added to RDEPENDS since it is only RRECOMMENDS of mender-update. Re-verified on hardware against Hosted Mender: successful A/B deployment + rollback of a broken artifact; the failure log shows mender-flash's block statistics, confirming it handles the write.

meta-mender-qcom/recipes-support/unoq-ab-integration/files/unoq-mender-persist
TheYoctoJester: Agreed — in a06630f this now seeds once, on first boot only (same as `device_type`). `/var/lib/mender/mender.conf` is the client's *fallback* config and `/etc/mender/mender.conf` (in the rootfs, refreshed by every A/B update) overrides it, so per-boot re-copying buys nothing — and the old behavior could clobber config set at runtime on /data.

PR #522, opened 2026-07-20
meta-mender-xilinx: add Mender integration for Digilent Arty Z7-20

TheYoctoJester:
Add a new SoC-oriented layer meta-mender-xilinx carrying Mender OTA integration for Xilinx / AMD Zynq boards, starting with the Digilent Arty Z7-20 (Zynq-7000 xc7z020). It integrates the u-boot-xlnx fork (meta-mender only appends to the plain u-boot recipe) with a rebased boot-code patch and a Zynq env_get_location MMC fix that is required for Mender's redundant raw-MMC environment on SD boot, plus an fpga-bitstream Mender update module that deploys PL bitstreams via the kernel FPGA manager with a boot-time reload service.

Board bring-up (machine, devicetree, U-Boot control DTS, platform-init) lives in the nested BSP sub-layer meta-arty-z7, kept separate so a plain bring-up build stays free of Mender patches. The board has no support upstream (no machine in meta-xilinx, no devicetree in mainline U-Boot), so the BSP is nested here for now.

A single kas build configuration (arty-z7-20, which installs the fpga-bitstream update module) and an experimental CI matrix entry are included.


danielskinstad (approved):
LGTM (I just did a quick skim)

PR #528, opened 2026-07-30
feat(tegra): add an opt-in Tegra-native rootfs update module

TheYoctoJester:
Mender's stock `rootfs-image` update module is written for u-boot and GRUB. On
Tegra it only works through a stack of adapters this layer maintains, and every
one of them papers over something the platform does not provide:

| adapter | what it papers over |
|---|---|
| `libubootenv-fake` | there is no u-boot environment, so `fw_setenv mender_boot_part` is a **no-op** and the module cannot switch slots at all |
| `ArtifactInstall_Leave_50_switch-rootfs` | the slot switch itself, done outside the module |
| `ArtifactCommit_Enter_50_verify-slot` | `nvbootctrl verify`, which the module has no way to call |
| `ArtifactRollback_Leave_50_abort-blupdate` | removing a staged capsule the module does not know it staged |
| `mender-update-verifier` | undoing state the shims cannot express |
| `RootfsPartA/B` in `mender.conf` | naming partitions the BSP layout already names `APP` and `APP_b` |

Everything the module needs is available natively, so this adds a Tegra specific
module that talks to the BSP directly: `nvbootctrl -t rootfs` for the running
slot and for verification, `/dev/disk/by-partlabel/APP` for the partitions,
`mender-flash` for the write, `oe4t-set-uefi-OSIndications` to arm the staged
capsule. No `fw_printenv`, no state scripts, no `mender.conf` partition entries,
no `upgrade_available`, and none of them are installed when the scheme is on.

## Opt in

```
TEGRA_MENDER_NATIVE_UPDATE = "1"
```

Off by default, because the existing scheme is verified and stays the default.
The payload type is what selects the module, so the two schemes cannot be mixed
within one artifact.

The image then builds a `.tegra-mender-native` artifact instead of a `.mender`
one, payload type `tegra-rootfs-image`, carrying the rootfs `.ext4` and
`tegra-bl.cap`. Shipping the capsule inside the artifact is what removes the
worst of the glue. It still provides `rootfs-image.version` and clears
`rootfs-image.*`, so nothing changes from the server's point of view.

## Three things worth knowing

**The capsule performs the slot switch, not `nvbootctrl`.** Applying a UEFI
capsule writes firmware into the inactive boot chain and makes it active, and
rootfs slots are paired to bootloader slots, so the rootfs follows.
`set-active-boot-slot` is used only to undo a switch during rollback; calling it
on the install path too would aim the capsule at the wrong chain.

**Slot verification needs a marker.** `RootfsRetryCountMax` is 1 here, so an
unverified slot does not survive a boot and something has to run `nvbootctrl
verify` on ordinary boots. It also has to stand down while an update awaits its
verdict, or the new slot is marked good before Mender commits and the A/B
rollback is gone. meta-tegra's `nv_update_verifier` keys that off
`upgrade_available`, which does not exist here, so it is disabled and
`tegra-rootfs-verify.service` takes over. The module writes a marker naming the
target slot, and the verifier stands down only while that slot is the one
running, so a marker left over from an interrupted deployment cannot condemn a
slot forever.

**Migration takes two deployments.** Modules are dispatched by the running
rootfs, so a device on the legacy scheme has no `tegra-rootfs-image` module yet:
first a legacy `.mender` artifact built from an image that contains the module,
then the native one. Same for later changes to the module itself.

Note that the stock `rootfs-image` module stays in the image on purpose. Without
`libubootenv-fake` it fails on the missing `fw_printenv` before writing
anything, which is the loud failure you want if someone deploys an ordinary
`rootfs-image` artifact here.

Two bits of build wiring look odd and are explained in comments: `IMAGE_FSTYPES`
uses `:append:tegra`, because kas configurations set `IMAGE_FSTYPES:tegra`
outright and a plain `+=` is discarded, and the flag is
`TEGRA_MENDER_NATIVE_UPDATE` rather than `MENDER_TEGRA_*`, because meta-mender
sanity checks `MENDER_` variables against a fixed list a layer cannot extend.

## Verification

Offline, against stubbed `nvbootctrl`, `mender-flash` and
`oe4t-set-uefi-OSIndications`: 42 assertions across 11 scenarios, passing under
busybox ash, dash and bash. Both shipped scripts parse under ash, since this
layer targets images such as `core-image-minimal` where bash is absent, the same
constraint #527 dealt with.

On hardware, a Jetson Orin NX 8GB (`p3768-0000-p3767-0001`, NVMe, JetPack 7,
L4T 39.2.0) running `core-image-minimal`:

| scenario | expected | result |
|---|---|---|
| plain update, commit | lands on the other slot, `nvbootctrl verify` run, no state scripts in the log | pass |
| `ArtifactInstall_Leave` fails after the capsule is staged | capsule removed, `OsIndications` disarmed, slot unchanged | pass |
| `ArtifactCommit_Enter` fails after the reboot | `set-active-boot-slot` back, deployment `failure`, board on the original slot | pass |

All of them in both slot directions. The broken artifacts are repacks of the
same build's `.ext4`, so the injected state script is the only variable.
Rollback issues an explicit `set-active-boot-slot` rather than trusting the
firmware's own fallback, which is why the A/B asymmetry noted in #527 does not
affect these paths.

## Known limitations, also in the README

**Every rootfs update reflashes the bootloader.** The capsule ships in every
artifact and is armed unconditionally, so each deployment writes the inactive
boot chain even when the firmware is identical, roughly 11 MB of QSPI on an Orin
NX. The legacy scheme arms it the same way, so this is inherited rather than a
regression. Fixing it means comparing firmware versions in `ArtifactInstall` and
skipping the arming, but the capsule is also what switches the slot, so the two
do not separate cleanly.

**Rollback does not roll back firmware.** Also true of the legacy scheme. The
part worth stating: the standby slot ends up holding new firmware next to a
rootfs that failed its commit, still flagged bootable, because the module clears
the unbootable flag once the write completes. Should the active chain later fail
to boot, the firmware falls back to a pair that was never validated together.

**The unbootable-rootfs fallback gap** applies exactly as it does to the legacy
scheme. A/B recovers a rootfs that boots but never commits, not one that fails
to boot.

Two nits, neither blocking: `disarm_capsule` zeroes all eight bytes of
`OsIndications` rather than only the capsule delivery bit, safe because nothing
else on this image sets those bits, and `tegra-rootfs-verify.service` uses
`ConditionPathExists=/usr/sbin/nvbootctrl`, so a moved `nvbootctrl` would make
the unit silently succeed and stop verifying.

### DIFF
--- a/meta-mender-tegra/README.md
+++ b/meta-mender-tegra/README.md
@@ -73,6 +73,106 @@ To solve this issue we create a new [custom partition layout](recipes-bsp/tegra-
 
 It is possible to auto-grow the UDA partition to fill remaining space with [this](https://gist.github.com/rishabnayak/a734d2720f43b8908e59564c14fa52e9) bbappend in a layer above `meta-mender-tegra`. It sets the UDA allocation attribute to `0x808`, removes partition id numbers, and moves the UDA partition to right before the `secondary_gpt` partition following [Nvidia documentation](https://docs.nvidia.com/jetson/archives/r35.6.0/DeveloperGuide/AR/BootArchitecture/PartitionConfiguration.html#partition-child-elements).
 
+## Tegra-native update scheme (opt in)
+
+By default this layer drives Mender's stock `rootfs-image` update module. That module
+is written for u-boot and GRUB systems, so on Tegra it only works through a stack of
+adapters: `libubootenv-fake` provides `fw_printenv`/`fw_setenv` (where
+`fw_setenv mender_boot_part` is a no-op), three Mender state scripts perform the
+actual slot switch outside the module, and `mender-update-verifier` exists to undo
+state the shims cannot express.
+
+Setting
+
+```
+TEGRA_MENDER_NATIVE_UPDATE = "1"
+```
+
+selects an alternative that talks to the BSP directly. It uses `nvbootctrl -t rootfs`
+for slot queries and verification, `/dev/disk/by-partlabel/APP{,_b}` for the
+partitions, `mender-flash` for the write, and `oe4t-set-uefi-OSIndications` to arm the
+capsule. None of the shims, state scripts or `mender.conf` partition entries are
+needed, and none are installed.
+
+### What changes
+
+The image builds a `.tegra-mender-native` artifact instead of a `.mender` one. It
+carries the payload type `tegra-rootfs-image` and two payload files, the rootfs
+`.ext4` and `tegra-bl.cap`. Shipping the capsule inside the artifact is what removes
+the worst of the glue, since the legacy `switch-rootfs` state script has to mount the
+freshly written slot just to fetch it.
+
+The artifact still provides `rootfs-image.version` and clears `rootfs-image.*`, so
+inventory and deployment logic look unchanged to the server.
+
+Note that the payload type is what selects the module, so the two schemes cannot be
+mixed within one artifact. The stock `rootfs-image` module is deliberately left in the
+image: without `libubootenv-fake` it fails immediately on the missing `fw_printenv`,
+which is the loud failure you want if an ordinary `rootfs-image` artifact is deployed
+to a device running this scheme.
+
+### Slot verification
+
+This platform sets `RootfsRetryCountMax` to 1, so a slot that is never verified does
+not survive a boot. Something has to run `nvbootctrl verify` on an ordinary boot, and
+it must stand down while an update is awaiting its verdict, or the new slot is marked
+good before Mender commits and the A/B rollback is gone.
+
+The legacy scheme keys that window off `upgrade_available` in the u-boot environment.
+The native scheme has no u-boot environment, so meta-tegra's `nv_update_verifier` is
+disabled and replaced by `tegra-rootfs-verify.service`. The update module writes
+`/var/lib/mender/tegra-update-pending` naming the slot the update is aiming at, and
+the verifier stands down only while that slot is the one running. A marker left behind
+by an interrupted deployment therefore cannot stop verification of some unrelated slot
+forever, which would condemn it.
+
+### Migrating an existing device
+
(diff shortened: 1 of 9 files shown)


TheYoctoJester:
Closing this in favour of a layer split, which is a better mechanism for the same
thing.

`TEGRA_MENDER_NATIVE_UPDATE` selected between two schemes that disagree on package
content and on `IMAGE_FSTYPES`, not merely on configuration, and expressing that as a
flag cost ten decision points across five files, three of them bbappends on meta-tegra
recipes. The schemes now live in their own layers, so the exclusion is structural: a
build takes `meta-mender-tegra-common`, exactly one of `meta-mender-tegra-classic` or
`meta-mender-tegra-native`, and a Jetpack layer, and inherits that scheme's class.

The update module itself is unchanged apart from moving into
`meta-mender-tegra-native`. The replacement is a stack of three branches on my fork:

- `tegra-layer-split`, the classic scheme carved out of the common layer, no native
  code in it at all
- `tegra-native-layer`, this module in its own layer, on top of the above
- `tegra-recovery-split`, the recovery system, on top of that

I will open them as three PRs once I have finished reviewing them locally.

Worth stating for whoever picks the first one up: the classic carve-out is a no-op for
existing builds. Against a pre-split reference build the `core-image-base` package
manifest is byte identical, 404 packages, and the only substantive differences in the
image's variable set are `INHERIT`, environment noise, and three search paths that gain
the new layer's empty directory. The native scheme and the recovery system are verified
on a Jetson Orin NX 8GB on NVMe, including an RCM flash that populates the recovery
partition byte for byte, followed by an over-the-air update on the flashed board.

TheYoctoJester:
The replacement is now open: #529 for the layer split, #530 for this module in its own layer on top of it. The recovery system follows in a third once its build configuration is in place.

PR #529, opened 2026-08-06
feat(tegra): carve the classic update scheme into its own layer

TheYoctoJester:
Rebased onto `wrynose` with #534 merged. #530 stacks on this one.

`meta-mender-tegra` has always held exactly one update scheme, so the scheme was simply what the layer did. Adding a second one calls for saying which is which, and a layer says it better than a variable: the two schemes disagree about which packages go into the image, not only about configuration, so layer selection makes them exclusive by construction rather than by a runtime check. #528 proposed the variable and it came to ten decision points across five files.

A build now takes three of these layers, the common one, exactly one scheme layer, and a Jetpack layer, and inherits the scheme's class:

```
INHERIT += "tegra-mender-classic"
```

`meta-mender-tegra-classic` holds the stock `rootfs-image` scheme: the three Mender state scripts, `libubootenv-fake`, `mender-update-verifier` and the `nv_update_verifier` wrapper installed onto `tegra-redundant-boot`. `meta-mender-tegra-common` keeps what neither scheme can do without, which is most of it: partition numbers, the storage device, the A/B size calculation, the flash layouts, the persistent machine-id, and #534's disk and ESP detection.

`tegra-mender-setup.bbclass` becomes `tegra-mender-common.bbclass`, and the rename is deliberate rather than cosmetic. A configuration still inheriting the old name would otherwise keep parsing and produce an image with no state scripts and no `fw_printenv` shims, which boots and can never update. Now it fails at config parse.

Two things had to move with their recipes rather than stay behind. The `PREFERRED_PROVIDER` block, all of it: one naming a recipe no layer provides is neither an error nor a warning, so leaving it in common would let a build without `libubootenv-fake` resolve `libubootenv` to oe-core's real implementation without saying a word. And `_MENDER_IMAGE_DEPS_EXTRA`, which deploys the state scripts for the artifact to bundle.

Selection is checked in the common class, the one layer present in every valid configuration. `BBLAYERS` and `INHERIT` are independent selectors and bitbake cannot express a conflict between layers, so nothing else catches both scheme layers being present with one of them inherited. That combination builds cleanly, since bbappends come from `BBFILES` and ignore `INHERIT` entirely, and it yields an image whose slot verification belongs to the other scheme. The check runs from a `ConfigParsed` handler rather than anonymous python, so the error is reported once, at configuration time, rather than once per recipe and blamed on whichever recipe parsed first.

Note that the classic layer's `tegra-state-scripts` now carries `DEPENDS = "tegra-mender-helpers"` across the layer boundary. It needs #534's library staged, not installed, because it splices it into each state script at build time; see that PR for why a state script cannot source anything out of the image.

`mender-uboot` stays in the common layer's `MENDER_FEATURES_ENABLE` although it is classic flavoured, because the kas configurations disable it globally and a disable beats an enable. Moving it would change what anyone who does not disable it gets. Noted in a comment as its own piece of work.

For an existing build this is a no-op, and that is measured rather than asserted: against a pre-split build of the same target `core-image-base.manifest` is byte identical at 404 packages, and 71 of 6950 task signatures change from three understood causes, none of which reach the image. Review with `--find-renames`, since 15 of the 25 files are moves into the new layer.

Exercised on an Orin NX 8GB on NVMe and on an Orin Nano 8GB devkit with a microSD and an NVMe both fitted: deployments in both slot directions, with the state scripts running under busybox ash on `core-image-minimal`, a rollback that condemned the target slot and the deployment that recovered it, and deployments against the other disk's ESP. Built for both Orin NX modules and parse checked for Thor and both Orin Nanos; the 16GB module has no hardware coverage since there is none on the bench.

The kas configurations in [mender-community-images](https://github.com/TheYoctoJester/mender-community-images) group by scheme and point at this branch while it is in flight, so they need re-pinning to the merge commit once this lands, and their floating tier is broken until then. #528 is closed in favour of this series.


TheYoctoJester:
Stacked on #534, which carries the two switch-rootfs fixes (ESP and slot resolution on the booted disk). Those two commits show up in this diff until #534 merges, at which point this branch gets rebased and they drop out.

PR #530, opened 2026-08-06
feat(tegra): add the Tegra-native update scheme as its own layer

TheYoctoJester:
Rebased onto `wrynose` now that #529 and #534 have landed, so the diff is three commits and nothing else. Replaces #528.

Mender's stock `rootfs-image` module is written for u-boot and GRUB systems, and on Tegra it only works through adapters `meta-mender-tegra-classic` maintains. `libubootenv-fake` fakes `fw_printenv`/`fw_setenv`, where `fw_setenv mender_boot_part` is a no-op, so the module cannot switch slots at all. Three state scripts do the real work outside the module, one of them mounting the freshly written slot just to copy a UEFI capsule out of it. And the module parses `RootfsPartA`/`RootfsPartB` out of `mender.conf` to find partitions the BSP already names.

Everything needed is available natively, so `tegra-rootfs-image` talks to the BSP directly:

```
nvbootctrl -t rootfs        which slot runs, and slot verification
/dev/disk/by-partlabel/APP  the rootfs partitions, named by the BSP layout
mender-flash                sparse-aware write to the passive slot
the UEFI capsule            the slot switch itself
```

Which device a label actually means, and which ESP the capsule belongs on, come from #534's `disk-helpers.sh`, sourced at runtime out of the rootfs since unlike a state script the module lives there. So the module carries no copy of that logic itself.

The capsule travels in the artifact alongside the rootfs, which is what removes the need to mount the freshly written slot just to fetch one. The artifact keeps the canonical `.mender` name and still provides `rootfs-image.version`, so nothing about uploading or deploying changes; only the payload type inside it differs, and `mender-artifact read` is how you tell the two apart.

`tegra-rootfs-verify` replaces the BSP's `nv_update_verifier`, which this scheme cannot use: that one keys the verification window off `upgrade_available` in a u-boot environment there is no longer any shim for. The replacement keys it off a marker naming the slot an update is aiming at, and stands down only while that slot is the one running, so a marker that outlives its deployment cannot disarm verification permanently. With `RootfsRetryCountMax` at 1 on this platform, a slot that is never verified does not survive a boot, so something has to run `nvbootctrl verify` on an ordinary boot and stand down while an update awaits its verdict.

Note that `nv_update_verifier` is disabled rather than removed. It arrives through `MACHINE_EXTRA_RDEPENDS` in meta-tegra's `tegra-common.inc` so it is in the image either way, and masking it breaks `tegra-redundant-boot`'s postinstall during the rootfs build. One thing only the split could reveal: before it, the classic layer's `FILESEXTRAPATHS` shadowed meta-tegra's unit with a shell wrapper, so a native build got the wrapper. With the classic layer gone the unit is meta-tegra's own and its `ExecStart` is `nvbootctrl verify` on every boot, which would mark a new slot good before Mender commits and spend the rollback window. So disabling it is load-bearing under the split, for a different reason than #528 gave.

The stock `rootfs-image` module is deliberately left in the image. Without `libubootenv-fake` it fails immediately on the missing `fw_printenv`, which is the loud failure you want if an ordinary `rootfs-image` artifact reaches a device running this scheme. That cuts both ways and there is no migration path: update modules are dispatched by the running rootfs, so a `tegra-rootfs-image` artifact on a classic device dies on the missing module under `/usr/share/mender/modules/v3`, and neither failure touches the slots. `meta-mender-tegra-native/README.md` says so, along with the scheme's limitations: every rootfs update reflashes the bootloader, a rollback does not roll back firmware, and the unbootable-rootfs gap that A/B does not cover.

Exercised on an Orin NX 8GB on NVMe and on an Orin Nano 8GB devkit with a microSD and an NVMe both fitted: the happy path in both slot directions, the post-reboot rollback in both directions, the pre-reboot rollback, and deployments against the other disk's ESP. Both directions matter, since UEFI treats the two slots asymmetrically. The Orin NX was then RCM flashed with this scheme's `tegraflash-tar` and took an update over the air, so flash, boot, update and commit are covered as one chain. Re-run on the Orin Nano after this rebase, with the classic scheme resolving against `wrynose` and only the native layer coming from the fork.

Build configurations are in [mender-community-images](https://github.com/TheYoctoJester/mender-community-images) under `tegra/jetpack7/native/`, for both Orin NX modules and both Orin Nano devkit variants. Only the 8GB modules have hardware coverage. The recovery system that builds on this scheme is held back for a separate pull request.

### DIFF
--- a/meta-mender-tegra/README.md
+++ b/meta-mender-tegra/README.md
@@ -46,6 +46,11 @@ release.
   The classic update scheme: Mender's stock `rootfs-image` update module, driven
   through this layer's state scripts and the `fw_printenv`/`fw_setenv` shims.
 
+- `meta-mender-tegra-native`
+  The Tegra-native update scheme: a `tegra-rootfs-image` update module driving
+  `nvbootctrl`, the BSP partlabels and the UEFI capsule directly, with none of
+  those adapters.
+
 - `meta-mender-tegra-jetpack7`
   Jetpack 7 specific parts, matching the `wrynose` branch of `meta-tegra`.
 
@@ -62,9 +67,18 @@ BBLAYERS += "\
 INHERIT += "tegra-mender-classic"
 ```
 
-`tegra-mender-common` is never inherited directly. The scheme class pulls it in,
-and on its own it would configure no scheme at all, so the common layer refuses
-that at parse time.
+For the native scheme, substitute `meta-mender-tegra-native` and
+`INHERIT += "tegra-mender-native"`.
+
+Two things are refused at parse time. Inheriting `tegra-mender-common` directly:
+the scheme class pulls it in, and on its own it would configure no scheme at all.
+And two scheme layers at once: the bbappends of both would apply whichever class
+was inherited, yielding an image whose slot verification belongs to the other
+scheme.
+
+The schemes are mutually exclusive on the device as well. Neither can install the
+other's artifact, so there is no migration path; see
+[meta-mender-tegra-native/README.md](meta-mender-tegra-native/README.md).
 
 ## Quick start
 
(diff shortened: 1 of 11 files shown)


TheYoctoJester:
Stacked on #534, which carries the two switch-rootfs fixes (ESP and slot resolution on the booted disk). Those two commits show up in this diff until #534 merges, at which point this branch gets rebased and they drop out.

PR #532, opened 2026-08-07
fix(tegra): default EXTRADEPS so non-tegra machines can parse

jolting:
mender-tegra.inc is pulled in via mender_%.bbappend, which applies to every machine while the layer is in bblayers. EXTRADEPS is only defined with a :tegra override, so on any non-tegra machine (e.g. a qemu VM flavor built from the same layer set) RDEPENDS:mender-update gets the literal string ${EXTRADEPS}, which is unbuildable. Provide an empty default.


jolting:
This will be superseded by #529, so closing PR.

PR #536, opened 2026-09-22
feat(nxp): add i.MX 93 11x11 LPDDR4X EVK support

TheYoctoJester:
Recreates `meta-mender-nxp` on `wrynose`, where the layer had been dropped, with Mender A/B for the NXP i.MX 93 11x11 LPDDR4X EVK booting from microSD. Only the EVK is carried over; `imx93-voipac` and `olimex-imx8mp-evb` are left out, since neither is verified and both spell the device tree variable `${fdt_file}`, which NXP's downstream U-Boot does not define.

No `meta-imx`, no downstream pins, no meta-mender fork. meta-freescale on `wrynose` is self-contained for i.MX93 and still selects NXP's BSP through `IMX_DEFAULT_BSP`. `MACHINE` is `imx93-11x11-lpddr4x-evk`, not NXP's consolidated `imx93evk`, which only requires the same machine configuration and adds other boards' device trees on top.

Defaults are keyed on the `mx9-generic-bsp` override meta-freescale already provides rather than on a machine list, and live in a `classes-global` class the image configuration inherits, so another i.MX 9 board picks them up without touching the layer. The BootROM offset is derived from `IMX_BOOT_SEEK` instead of restated per board, and `boot.cmd` holds nothing board-specific and is shared. The i.MX 8 families would take the same defaults unchanged, but they are left alone until somebody has booted one.

Three patches are carried. Two are version skew: oe-core `wrynose` ships u-boot 2026.01 and meta-mender is written against it, while meta-freescale pins `u-boot-imx` to lf_v2025.04, so meta-mender's `0002` misses its `env_default.h` hunk and `config_mender.h` asks for a `CONFIG_ENV_REDUNDANT` that tree does not have. Both go when `u-boot-imx` rebases past the rename. The third disables EFI capsule authentication, without which the DTB build wants `cert-to-efi-sig-list` from `meta-imx-sdk` and a meta-imx-free build cannot compile U-Boot at all. Note that this one is a meta-freescale gap rather than a Mender concern, and sits here only because this is the layer the board already pulls in. Happy to move or drop it.

Build configurations are in [mender-community-images](https://github.com/TheYoctoJester/mender-community-images) under `yocto/wrynose/{tagged,floating}/`, since `wrynose` moved the kas files out of here. They pin this layer from my fork for now.

Two things I would welcome a view on: where the EFI capsule fragment belongs, and whether this repository should carry the two refreshed U-Boot patches at all.

---

Written with Claude Code (Opus 5), which produced the layer and the build configurations, drove the builds, and ran the bench work against the board over its serial console, SD-mux and Hosted Mender. The commits carry it as co-author.


TheYoctoJester:
---

**Update:** the layer is no longer mx9-only by construction.

Its defaults are now keyed on the `mx8-generic-bsp` / `mx9-generic-bsp`
family overrides meta-freescale already provides, rather than on a machine
list. Before this, an i.MX 8 board picking the layer up silently inherited
meta-mender's GRUB/UEFI defaults and produced an unbootable image.

Three things moved out of board configs and into the layer, because they are
properties of the SoC family:

- `MENDER_IMAGE_BOOTLOADER_FILE`, and the partition alignment that has to
  clear `imx-boot`
- `boot.scr` appended to `IMAGE_BOOT_FILES`
- `MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET`, now **derived** from
  meta-freescale's `IMX_BOOT_SEEK` instead of restated. The two differ in
  units and are not a fixed pair — `IMX_BOOT_SEEK` is 32 KiB on most parts
  but 33 on the i.MX 8M Mini — and a wrong value yields a board with no
  serial output at all, so it is worth nobody computing it by hand.

`boot.cmd` turned out to contain nothing board-specific and is now shared,
with machine-named subdirectories still available as an override.
`u-boot-scr`'s `COMPATIBLE_MACHINE` and the `DEPENDS` on it are keyed on the
same condition, which removes the failure mode where the two disagree.

Resolved for `imx93-11x11-lpddr4x-{evk,frdm}`, `imx8mp-lpddr4-evk` and
`imx8mm-lpddr4-evk`: all four select `mender-uboot` with `mender-image-sd`,
and the Mini alone derives 66 sectors. A full rebuild of the verified EVK
image produces an identical partition layout, `fw_env.config` and boot
container offset, so this is behaviour-preserving for the board that was
tested on hardware.

Adding a board in a covered family is now a defconfig patch, optionally a
device tree name, and a storage device. i.MX 6 and 7 remain out of scope:
32-bit, mainline BSP, `u-boot-fslc` rather than `u-boot-imx`, so they need a
second bbappend and a `bootz` script.

oldgiova (approved):

